PR #52, opened 2015-11-04
fixes/Roman Numeral Converter Infinite Loop Issue

SaintPeter:
Fixes FreeCodeCamp/FreeCodeCamp/issues/3991


ltegman:
Woohoo -- I was just hoping someone else would do this soon so I wouldn't have too! :stuck_out_tongue_closed_eyes:  Love the solution :smile:

PR #66, opened 2015-11-08
Update bonfire-repeat-a-string-repeat-a-string.md

shadowfool:
Removed solution that used str.prototype.repeat (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/repeat) function as it defeats the purpose of this exercise (to define your own repeat function.) Moreover, the explanation said that there's a loop when in reality there is no loop in the code. I felt this was misleading and spreading misinformation on how to solve this problem. Perhaps the problem should restrict the use of str.prototype.repeat

Added a recursive solution which better exemplifies what the third solution was attempting to show.

### DIFF
--- a/bonfire-repeat-a-string-repeat-a-string.md
+++ b/bonfire-repeat-a-string-repeat-a-string.md
@@ -44,31 +44,27 @@ function repeat(str, num) {
 repeat("abc", 3, "");
 ```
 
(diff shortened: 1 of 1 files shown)


Rafase282:
Great work, thanks @shadowfool

PR #73, opened 2015-11-10
More

Rafase282:


Rafase282:
I ended up adding it directly to the repo after I though I had merged this, so no need for this PR anymore.

PR #77, opened 2015-11-12
no need to numCopy or have +decimalValue

benschac:


SaintPeter:
The content of this change seems to be adding your name to the thanks?  Did you previously make a change to this Wiki?

PR #84, opened 2015-11-17
added a shorter solution

sabahang:
added a solution using direct array manipulation


SaintPeter:
I like your solution, but you should add your own code explanation section. 

Example here:
https://github.com/FreeCodeCamp/FreeCodeCamp/wiki/bonfire-mutations

sabahang:
@SaintPeter thanks for comment, I added an explanation section for my solution although there wasn't much to explain as the code comments should suffice

PR #86, opened 2015-11-18
Coded change

coded9:
To check for truth values we must first check whether the property exists or not and the property has a truth value.(Corrected from || to && )


Rafase282:
Already fixed with another PR

PR #87, opened 2015-11-18
fix code to pass tests

HermanFassett:
Fixing code so that it will pass the tests as in issue https://github.com/FreeCodeCamp/FreeCodeCamp/issues/4505


SaintPeter:
Don't you need to check if `collection.hasOwnProperty[pre]` first?  Otherwise I think it errors out for some tests.

HermanFassett:
You don't need to but probably should. I'm lazy and didn't put it and should have. Maybe test should make it so you have to?

SaintPeter:
I don't know if it's possible for the test to test for it.  You should put it in, though, because "in the real world" it would case a runtime error.

PR #93, opened 2015-11-18
Using some method

coded9:


SaintPeter:
This is not a correct usage of `.some()` - the returned value is not being used.  Additionally, I think it may exit early.

PR #101, opened 2015-11-19
Added my solution to bonfire sorted union wiki

sabahang:
Added a solution that uses filter to tackle duplicate elements


SaintPeter:
Thanks! :+1:

PR #106, opened 2015-11-20
Splits campsites by countries and then sorts them

bugron:
Closes FreeCodeCamp/FreeCodeCamp#4282.


SaintPeter:
Would there also be interest in further sub-dividing these by state/province/political subdivision?  At least we could set a standard with the US states.

bugron:
I'm currently finishing the rest of cities and after merging you can do that if you want.

SaintPeter:
@bugron Will do, thanks!

bugron:
OK, ready for the QA.

SaintPeter:
FAN-tastic!

PR #115, opened 2015-11-22
New Solution for Bonfire-Sorted Union using reduce()

coded9:


SaintPeter:
You need more explanation about how your reduce flattens and your filter removes.  Also, the code explanation below your solution doesn't seem to match your solution.

coded9:
It wasn't my explanation as soon as I made a pull request,someone made another pull request my explanation was made in the comments itself.

QuincyLarson:
I merged your other solution that used Reduce.

PR #116, opened 2015-11-22
Update solution with an answer using reduce.

benschac:
Example with Array.reduce() method.  It's not as elegant as the second users solution but it does use the recommended reduce method.

### DIFF
--- a/Bonfire-Sorted-Union.md
+++ b/Bonfire-Sorted-Union.md
@@ -68,7 +68,27 @@ function unite() {
 - we use `filter` to remove the duplicate elements by checking the index of each element and removing same elements with different positions
 - ordering will be preserved as we didn't mess with it
 
+#### Third Solution:
+
+```js
+function unite(arr1, arr2, arr3) {
+  var args = Array.prototype.slice.call(arguments, 0), arr = [];
+args.reduce(function(curr, next) {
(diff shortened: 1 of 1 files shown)


QuincyLarson:
Hi @benschac we also got another solution using Reduce, and I ended up merging that first. Please take a look at it and if you think this solution is superior, feel free to re-open this pull request. Thanks for your contributions :)

benschac:
To be honest with you I do feel like me is a little more concise and easier to read for a couple of reasons:
-  naming conventions stick with documentation standards that are common.
-  when starting to learn these functional basics the logic of this is a little easier to grasp on to. First you flatten (with concat), then you filter that flattened array by method chaining. The logic is much easier to follow.
- I use the 4th param when starting reduce with an empty array which I think is more expressive and clear to someone reading the code that reduce is starting off with an empty array.
- I don't use a triple === when not needed.


-- The last (and hopefully my best argument :)) My solution has a lot less steps for the computer to take than the version you accepted.  The other solution:
- calls filter for each elements concat.
- checks the WHOLE reduced array each time for the indexOf() that element.

My solution:
- Calls filter once.
- Pushes each unique element into an array, so you NEVER have to check against the entire array each time.

PR #118, opened 2015-11-23
New solution for Bonfire Arguments optional

coded9:


QuincyLarson:
Thanks for adding this alternate solution! :)

PR #120, opened 2015-11-23
New Solution for Bonfire Roman Numeral Converter

coded9:


SaintPeter:
This solution need to have the `!` versions of letter removed and otherwise be optimized.  You can remove a significant number of the values in your arrays and still have it work properly.

Please rework this and re-submit.  

Thanks and happy coding!

PR #122, opened 2015-11-23
Added my solution and removed an incorrect solution for bonfire drop it

sabahang:
- The removed solution was wrong as it used `filter()` and removed all the elements that didn't satisfy the condition rather than just the ones that don't satisfy the condition before the index of the first found element
- Added my solution to the problem 
- fixed formatting


Rafase282:
Does that means that @mhelmetag was incorrect? If so, and given that it was removed, edited then his name should be removed until he has a chance to fix it if no solution was provided using his method no?

sabahang:
yap, waiting on his input...

mhelmetag:
@Rafase282 and @sabahang,

I have no idea how my attempt got added to this list of solutions...

I looked at my most recent attempt (and correct solution) and that was just looping through the array and using `Array.shift`. 

I think it would be fine to remove my (incorrect) attempt and name from the wiki. No big deal :)

Rafase282:
Alright, thanks for the input @mhelmetag 
@sabahang as soon as you make the changes I'll merge the PR.

sabahang:
@Rafase282 removed @mhelmetag from the contributors list. Thanks both for your cooperation

PR #126, opened 2015-11-27
Fixes the word Argument in link

chrisipanaque:
Fixes the word Argument in text link, where it points to more
information about optimization issues and managing arguments.


bugron:
@chrisipanaque thanks for the fix. Verified and merged!

PR #135, opened 2015-12-05
R282 copy

Rafase282:
As users were asking about the second code, I added some explanation and links to make it easier to understand.


Rafase282:
Well it shows more commits than I expected. Anyways, feedback on the new explanation?

PR #137, opened 2015-12-05
Organizing United States Cities by State

zenware:
I also found some Canadian Cities that were incorrectly categorized as US Cities.
Hampton Roads and University Center were ambiguous to me even with google searches, if anyone could clear up where they belong that would be great.


SaintPeter:
This is great!  
There was another recent issues that noted that both San Francisco and San Jose are the California versions, not in Costa Rica.  Do you think you could make that change as well?

PR #138, opened 2015-12-06
Deleted Argentina - Parana url link

bogadodiegoh:
The url link for Argentina - Parana link is the same as Brazil -
Apucarana url link.
"Argentina - Parana" group does not exist , so the correct url is the Brazil - Apucarana.


SaintPeter:
:+1:

PR #139, opened 2015-12-06
Changed Santa Rosa link from Argentina to United States

bogadodiegoh:
The Santa Rosa group belongs to Santa Rosa, California in United States.


SaintPeter:
There was a change made to break out the file by States.  We definitely need this fix, though.  Can you please update?

bogadodiegoh:
Updated with PR #140.

PR #151, opened 2015-12-14
Third solution for Bonfire Check Palindromes

coded9:


Bonfire-Check-for-Palindromes.md
```diff
@@ -49,6 +49,18 @@ function palindrome(str) {
 # Code Explanation:
 We use regular expressions to replace any uppercase letter into a lowercase. Then we check if the string is the same as the reversed string using `split()` to split the original string by characters, then reverse and then join it back together.
 
+##Third Solution
+```js
+function palindrome(str) {
+  //Remove everything in the string except alphabets and numbers
+var strStrip = str.replace(/[^A-Za-z0-9]/g,"").toLowerCase();
+//Check for palindrome condition 
+  if(strStrip.split("").reverse().join("")==strStrip)
```
Rafase282: `return strStrip.split("").reverse().join("")==strStrip` will also handle the returning the boolean. You don't have to add it but just letting you know.

SaintPeter:
I don't see anything in this new solution that isn't in the prior two.  Additionally, there no code explanation and only minimal comments.

Rafase282:
Thanks @SaintPeter That is something I was about to check. After a third solution, new ones get a hard time to introduce new things.

Bonfire-Check-for-Palindromes.md
coded9: Thank you ;)

PR #154, opened 2015-12-15
Another solution for Bonfire Inventory update

coded9:


SaintPeter:
This solution looks very similar to the first example solution.

What is it that you feel you are bringing that is new to this problem?  What is different from what has come before?  

The main difference between the first and second solutions is their use of funcitonal programming.   That seems worthwhile to have.

coded9:
I didn't see any of the solutions indeed :( Sorry :(

coded9:
It seems mine is a bit different :o:

SaintPeter:
A bit different, sure - but it's more like the natural variation from programmer to programmer.

I've seen you make several of these PRs and I'm going to ask that if you do them in the future that you have a clear vision for how the solution you are adding gives value to a new programmer.  Are you doing something in a new or unusual way?  Are you using a function or structure that has not been used before.  

Essentially, what do your fellow campers get out of reviewing *your* solution above and beyond the existing solutions?

I'm going to close this one because I don't think it adds enough to warrant its inclusion.

coded9:
I didn't use any function or something,but I solved in the  simplest way possible

PR #156, opened 2015-12-16
Update Bonfire-Missing-letters.md

Narshe1412:
Added another take in Solution 2, using two index variables in the For--Loop. This is not something often seen in FCC exercises but may be helpful for future takes.


SaintPeter:
This alternate solution is actually more complex and harder to read.  I don't see how it adds to user knowledge.

PR #158, opened 2015-12-17
Add clause about bots on FCC

BerkeleyTrue:


SaintPeter:
I would say "without prior approval from FCC staff".  We've got Purdybot in the /casual channel.

BerkeleyTrue:
I thought `at our discretion` would grandfather him in. I don't want other campers asking us for permission  to enable bots.

SaintPeter:
It makes it sound like not bot ever will be allowed, which is clearly not true.  Only the banning would be "at our discretion".

Here is what I would say:
> No bots are allowed in any of the Official Chat Rooms without prior explicit permission from the FCC Core Team.

Don't mention banning - that's implied.  Just state that we don't allow them unless we say it's ok up front.  We have expressly allowed Purdybot, so that's good.  We want people to ask before they do a bot and we expect that we'll probably say no, but all the rules need to say is that they have to ask first.  How we choose to use this rule is entirely at our discretion.  It gives us flexibility without tying our hands.

SaintPeter:
IMO, it's a net positive to have campers interested in working on bots.  If they ask, or they have a feature idea, we can redirect them to work on camperbot.  I agree, we don't want primitive bots cluttering up the official channels, but neither do we want to do a hard shutdown on folks who are interested.  We need to capture that interest and put it to good use.

BerkeleyTrue:
This isn't to discourage people from working on bots. But they shouldn't use FCC as their playground. I don't want to give the impression that we allow some bots because I don't think we should. This doesn't stop people from using their bots in their own or other org chatrooms.

SaintPeter:
Well, we've already given the impression that we allow it because we do: purdybot.  I quite enjoy purdybot in /casual.

I stand by my suggested language above.   At a bare minimum I would remove the reference to banning, since it is implicit in the presence of the requirement in the CoC.

BerkeleyTrue:
You make some valid points. @SaintPeter I'll update the copy with your suggestions.

Rafase282:
When and why did I closed this?

SaintPeter:
@Rafase282 When you accidentally deleted `master` that closed any open PRs against it.

Rafase282:
@SaintPeter ah okay, I though I might got my account hacked or something.

Rafase282:
What's the status on this one?

BerkeleyTrue:
Updated

BerkeleyTrue:
@Rafase282 :shipit:

PR #159, opened 2015-12-17
Update Waypoint-Change-Text-Inside-an-Element-Using-jQuery.md

freeradik:
Proposing another solution to the waypoint.


SaintPeter:
Thanks, nice observation.

PR #161, opened 2015-12-17
Added regex based solution

crisvdkooij:


SaintPeter:
Clever solution.  However, it could use some whitespace/formatting to make it more readable.  We're not doing these "code golf" style.  Can you please break it across several lines to make it more understandable?

crisvdkooij:
You're right. Now it's much more readable. :)

SaintPeter:
Much better!  Thanks!

PR #165, opened 2015-12-19
Removed outdated solution

crisvdkooij:
The second solution in the file didn't pass the new tests


Rafase282:
Now that I take a look at it, the regex is not explained in the code solution for the short version. Can you do that?

crisvdkooij:
Allright, I have added explanations and reformatted to more closely follow the wiki template.

PR #166, opened 2015-12-19
Add Javascript Allonge Six Book

SaintPeter:
As recommended by @terakilobyte


Rafase282:
Already bookmarked it!

PR #167, opened 2015-12-20
added a new solution with spread operator to get the max and min of an array

aganita:


Rafase282:
It would be good if you add explanation and link to documentation for the campers. Maybe just a section for that instead of modifying the existing solution?

aganita:
you have a point. Let me add my version as one of the possible solutions.

PR #168, opened 2015-12-20
4th solution - using while instead of for loop

MikeRWood:


Rafase282:
@MikeRWood you need to add the code the proper way with the ''' there. Check the wiki on formatting or just look at the previous examples.

Also I think the second and third example should he merged, the first is done without any special functions, the second with a sepecualized function, and the third with a while loop.

PR #171, opened 2015-12-21
added more efficient code for roman numeral converter

aganita:


SaintPeter:
While this very well may be a more efficient solution, I don't believe the differences are enough to warrant a whole second solution.  If there was a fundamental difference in the solution - use of alternate functions or functional programming, then right on, but we don't need to clutter up the Wiki with substantially similar solutions.

Thanks for submitting and happy coding!

PR #174, opened 2015-12-21
Add warning message

abhisekp:
Preview: https://github.com/abhisekp/wiki/blob/fix/moderator-policy/Gitter-Moderation-Policy.md

Ref: Based on [@SaintPeter's Reply](https://gitter.im/FreeCodeCamp/admin?at=5678479a40f33d527af766a9) in admin room.

cc: @SaintPeter

### DIFF
--- a/Gitter-Moderation-Policy.md
+++ b/Gitter-Moderation-Policy.md
@@ -7,7 +7,7 @@ People who come across as excessively negative
 
 2) Moderators will then send them this message:
 
-```
+```md
 Hey, I'm sorry but I had to temporarily ban you.
 
 I can consider unbanning you, but I need you to do something first.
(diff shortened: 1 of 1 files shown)


Rafase282:
Looks good to me.

SaintPeter:
I'd like to get @QuincyLarson's feedback, since this is global FCC policy.  I don't mind being an examplar, but I want to make sure that everyone is on board.

QuincyLarson:
I think that it would have been just as effective to send him the canned "why do you think I banned you" question - maybe more effective. Here's why: It forces the person to actually stop and think, and doesn't give them any clear target for argumentation. If someone's angry and wants to fight you, one of the best ways you can calm them down is to ask them a bunch of open-ended questions. And I think that applies here.

Also, I think we should all follow a standard procedure. This way, we can ensure everyone is treated fairly. This is the purpose of such a system.

Finally, I think following this procedure will be much easier than trying to come up with a "lecture" on a case-by-case basis. So I argue that we should reject this pull request.

SaintPeter:
While I agree that the instructions are probably good for policy, in my personal experience (based on ~10 years running a gaming guild), banning first then warning can risk escalating the situation, especially if the conduct is not overtly against the CoC or marginal.  A polite contact can de-escalate the situation and actually build loyalty over the long term.

To be clear - the post policy should remain clear - but I do believe we need to rely on the judgement of our moderators to determine when an immediate ban is appropriate vs. a casual engagement.  These rules should be guidelines and cover for administrative action, not a straitjacket.

SaintPeter:
As per Quincy, closing this.  No change to the official policy.

PR #181, opened 2015-12-24
Binary Agents Solution 4

CaroleAnneHannon:
Using replace and regular expressions


Bonfire-Binary-Agents.md
```diff
@@ -95,6 +95,22 @@ function binaryAgent(str) {
...
+	return str.replace(/\d+\s?/g,parse);
+}
+```
+
+# Code Explanation
+- First we make an inline function to convert a single byte into an integer, then into a character
+- We use `replace()` with RegExp to capture a sequence of digits and an optional space (optional because the last sequence won't have a space after)
+- the replace uses the inline function on each match
+- the conversion has to use an inline function because the results are being altered before being returned (see: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace for more details)
+
```
Rafase282: Please use [ ] ( ) for links. No space though.

Rafase282:
You forgot to add your username.

PR #182, opened 2015-12-25
added third solution for search and replace problem

aganita:


Rafase282:
Could you add an explanation of your code? It is different from the first example, but it uses regex like the second one along with helper functions. many campers are not able to figure out the functionality by just reading the code, some can't even read the code and get a gist of it. The center part of your code does not even have comments for a brief explanation.

aganita:
got it. I have updated the comments. how does it look now?

Rafase282:
That's better, thanks.

PR #197, opened 2015-12-28
Add reference to bonfires when sending thanks (mass)

hallaathrad:
Mass update.


Rafase282:
There are conflicts with this branch.

PR #214, opened 2016-01-01
add new articles

0x0936:
This article explains how to log in to a local version of FCC with a user's GitHub account. Added an img/ folder to root with subfolder of 0x0936/ to host images for this article.


SaintPeter:
I really like this!  Well written, well illustrated.

One concern, as mentioned in Gitter Chat, we need to resolve where images should go.
As @ddaems suggests:
`/images/How-To-Login-To-Your-Local-FCC-Site/imagename.png`

PR #215, opened 2016-01-02
add new article

0x0936:
This article attempts to explain how a user may log into their local FCC site using their GitHub account. Also, added root folder images/ with subfolder <article name>/ to organize images for this and future articles.


SaintPeter:
This is fantastic!  Thank you!

PR #226, opened 2016-01-03
waypoints 94 and 95

CaroleAnneHannon:


Rafase282:
I forgot to mention that for this one too. Please add the 'js' code tag

PR #240, opened 2016-01-06
update contact info for new campsites to mine (instead of quincy).

hallaathrad:
Removed list of cities as the were already added in #238


Rafase282:
It looks like it was @QuincyLarson who added your name so it should be safe to merge, confirmation?

hallaathrad:
Yes. I would've merged it myself, but wanted to remove the list of cities (already updated yesterday).
Since I made an alteration to his commit, you're the safeguard before merging.

It's safe to merge if you approve.

PR #245, opened 2016-01-07
Added wiki for Bonfire: Caesar's Cipher

anuragaryan:
Please review.

### DIFF
--- a/Bonfire-Caesars-Cipher.md
+++ b/Bonfire-Caesars-Cipher.md
@@ -1,0 +1,68 @@
+# Problem Explanation:
+- You need to write a function, which will take string encoded with
+*Caesar cipher* as a parameter and decode it.
+- The one used here is ROT13 where the value of the letter is
+shifted by 13 places.
+e.g. 'A' ↔ 'N', 'T' ↔ 'G'.
+- You have to shift it back 13 positions, such that 'N' ↔ 'A'.
+
+## Hint: 1
+Use *String.charCodeAt()* to convert the English character to ASCII.
+
+
+## Hint: 2
+Use *String.fromCharCode()* to convert ASCII to English character.
+
+## Hint: 3
+Leave anything that doesn't come between A-Z as it is.
+
+## Spoiler Alert!
+[![687474703a2f2f7777772e796f75726472756d2e636f6d2f796f75726472756d2f696d616765732f323030372f31302f31302f7265645f7761726e696e675f7369676e5f322e676966.gif](https://files.gitter.im/FreeCodeCamp/Wiki/nlOm/thumb/687474703a2f2f7777772e796f75726472756d2e636f6d2f796f75726472756d2f696d616765732f323030372f31302f31302f7265645f7761726e696e675f7369676e5f322e676966.gif)](https://files.gitter.im/FreeCodeCamp/Wiki/nlOm/687474703a2f2f7777772e796f75726472756d2e636f6d2f796f75726472756d2f696d616765732f323030372f31302f31302f7265645f7761726e696e675f7369676e5f322e676966.gif)
+
+**Solution ahead!**
+
+## Code Solution:
+#### First solution
+
+```js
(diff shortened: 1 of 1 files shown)


Rafase282:
HI, Could you please elaborate more on the code explanation on the right section. You give a nice overview but the internals could be very confusing for people starting. I myself had to slow down and read again to get the gist of the inner workings and while I'm not a pro by any means I think I have gone beyond the fresh camper status.

SaintPeter:
Also, please make your code less compact.  You need to ALWAYS use curly braces and have good whitespace between operators.   We are NOT playing code golf.  Good code is human readable.

IE:
```js
 if(str.charCodeAt(i) < 78) {
     nstr += String.fromCharCode(str.charCodeAt(i)+13);
 } else {
      // Shift the character 13 places backward
     nstr += String.fromCharCode( str.charCodeAt(i) - 13);
}
```

anuragaryan:
@Rafase282 Added explanation. Check if it's okay.
@hallaathrad Done. :+1: 
@SaintPeter Thank you for the advice. I'll remember that.

Rafase282:
@anuragaryan Looks good to me, the `continue` is what threw me off but by the name I get an idea I would recommend you add [continue](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/continue) in there so users can read up on it.

Also please follow this instructions for [squashing your commits into one.](https://github.com/FreeCodeCamp/FreeCodeCamp/wiki/git-rebase#squashing-multiple-commits-into-one)

hallaathrad:
hahaha my solution to this one was so dirty I'm actually going to post it here so you can flame and hate.

```js
function rot13(encodedStr) {
  var codeArr = encodedStr.split("");  // String to Array
  var decodedArr = []; // Your Result goes here
  // Only change code below this line
  decodedArr=codeArr.map(function(letter){
    switch(letter){
      case 'N':return 'A';
      case 'O':return 'B';
      case 'P':return 'C';
      case 'Q':return 'D';
      case 'R':return 'E';
      case 'S':return 'F';
      case 'T':return 'G';
      case 'U':return 'H';
      case 'V':return 'I';
      case 'W':return 'J';
      case 'X':return 'K';
      case 'Y':return 'L';
      case 'Z':return 'M';
      case 'A':return 'N';
      case 'B':return 'O';
      case 'C':return 'P';
      case 'D':return 'Q';
      case 'E':return 'R';
      case 'F':return 'S';
      case 'G':return 'T';
      case 'H':return 'U';
      case 'I':return 'V';
      case 'J':return 'W';
      case 'K':return 'X';
      case 'L':return 'Y';
      case 'M':return 'Z';
      default: return letter;}
  });
  // Only change code above this line
  return decodedArr.join(""); // Array to String
}

rot13("SERR PBQR PNZC");
```

Please feel free to insult me. haha

Rafase282:
@hallaathrad I havent done mine but I was going to go for the solution using math instead of yours. That is the "official" solution as show by the creator on the live stream demo and I knew right away that I had to avoid it. Still yours gets the job done so no issues.

hallaathrad:
haha You're too kind @Rafase282. :+1:

anuragaryan:
@hallaathrad you need to be so zen to write that code. :+1: 
@Rafase282 can we merge it now? :laughing:

Rafase282:
Thanks @anuragaryan

SaintPeter:
@hallaathrad That was basically my proposed solution and I wrote the challenge.

I used a lookup object instead.  I actually used Excel to generate the code, using concatenation in Excel formulas.

hallaathrad:
@anuragaryan Funny you mention that. :+1: I'm all about Zen. 
My quote generator was on zen quotes. codepen.io/hallaathrad/pen/WrwWoK
I'll take your words as a compliment haha

@SaintPeter Good to know! :D If you came up with a similar approach I must not be that far off from the right path. Thank you as well.

PR #246, opened 2016-01-08
Created Checkpoint - Counting Cards

leftynaut:
Version 1 of Checkpoint - Counting Cards

### DIFF
--- a/Checkpoint-Counting-Cards.md
+++ b/Checkpoint-Counting-Cards.md
@@ -10,11 +10,16 @@ Having more high cards remaining in the deck favors the player. Each card is ass
 | ------ |:----------------------:|
 | +1     | 2, 3, 4, 5, 6          |
 | 0      | 7, 8, 9                |
-| -1     | 	10, 'J', 'Q', 'K','A' |
+| -1     | 10, 'J', 'Q', 'K','A'  |
 
 ***Instructions***
 
-You will write a card counting function. It will receive a card parameter and increment or decrement the global count variable according to the card's value (see table). The function will then return the current count and the string "Bet" if the count is positive, or "Hold" if the count is zero or negative.
+You will write a card counting function. It will receive a **card** parameter and increment or decrement the global **count** variable according to the card's value (see table). The function will then return the current count and the string **"Bet"** if the count is positive, or **"Hold"** if the count is zero or negative.
+
+**Example Output:**
+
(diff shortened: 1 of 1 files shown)


Checkpoint-Counting-Cards.md
```diff
@@ -0,0 +1,105 @@
...
+- If the final count is 0 or less, return `# Bet`
+
+## Hint: 1
+- Use a `switch` (or `else if`) statement to count the value of each card.
+
+## Hint: 2
+- Add/subtract the value of each card to variable `count`. If the card is worth 0, don't do anything!
+
+## Hint: 3
+- After you've counted the cards, use an `else` statement to check the value of `count`.
```
SaintPeter: This is wrong.  You need to use an `if` statement to check it.

SaintPeter:
Aside from the minor error this looks pretty good.  You might want to add a hint or a note about making sure there is a space between the count and the word.

leftynaut:
Fixed. See new commit.

hallaathrad:
I've noticed this is not the first PR in which this happens, and forgive me if I'm missing something but wasn't it Waypoint instead of Checkpoint?
Seeing it for a second time is actually making me doubt it.

leftynaut:
Not according to the [map](http://www.freecodecamp.com/map)

Checkpoint-Counting-Cards.md
```diff
@@ -0,0 +1,111 @@
+# Checkpoint: Counting Cards
+
+***About Blackjacks***
```
SaintPeter: The game is called "Blackjack", singular.

Checkpoint-Counting-Cards.md
```diff
@@ -0,0 +1,111 @@
...
+    case 3:
+    case 4:
+    case 5:
+    case 6:
+      count++;
+      break;
+    case 7:
+    case 8:
+    case 9:
+      break;
```
SaintPeter: You can technically omit these lines entirely (66-69) because they serve no function.  If not match is found in a switch statement then nothing happens.

SaintPeter:
Definitely improved.  Two more minor edits.  Sorry I didn't catch these the first time around.

leftynaut:
Fixed. See new commit.

SaintPeter:
Fab!  Thank you!

PR #248, opened 2016-01-08
Cleaning crew, Add declarative solution.

hallaathrad:
Wanted to synthesise the solutions, but I'm not sure if the different examples actually do give insights on methodically different albeit apparently similar ways.
What do you think, @Rafase282? Help me come up with a `.reduce` version so we don't have to use additional variables, initialise it with the starting char code and return directly? anyone?

----
Assigned to self to remind me I want to figure this one out :+1:

### DIFF
--- a/Bonfire-Missing-letters.md
+++ b/Bonfire-Missing-letters.md
@@ -16,7 +16,7 @@
 **Solution ahead!**
 
 ## Code Solution:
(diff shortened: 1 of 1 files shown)


Rafase282:
@hallaathrad hit me up on gitter.

PR #251, opened 2016-01-10
Waypoint heroku deploy

eferdman:


SaintPeter:
:+1: 
A solid start. We can add some more images later.

PR #258, opened 2016-01-12
Checkpoint profile lookup

leftynaut:
Created: Checkpoint - Profile Lookup


Rafase282:
Hi, looks good could you try to send it without all the merge commits? Also if it is a single commit that would be best. Feel free to read [this](https://github.com/FreeCodeCamp/FreeCodeCamp/wiki/git-rebase#squashing-multiple-commits-into-one)

leftynaut:
Thanks for the link, not sure how I keep doing that. Hopefully this explains it.

PR #261, opened 2016-01-12
Updated Map

leftynaut:
Added one link and one placeholder


Rafase282:
@leftynaut Hi, could you tell me what you changed, I noticed line 201 the new placeholder but there is the other extra number comes from? it went from 103 to 105. Sorry is hard to tell so I figured I would ask. other than that it seems good to merge, just want to check what is being added first.

Rafase282:
Never mind, figured out how to get rich diff.

leftynaut:
Added lines 201 and 230. Had to increase the numbers from 201 on.

PR #263, opened 2016-01-12
Created Article Checkpoint-Profile-Lookup

osterbergmarcus:
Created article for the checkpoint profile lookup


Rafase282:
There are some conflicts and also can you try to do it all in one commit when you try again?

PR #270, opened 2016-01-15
Added Solution #2 to Bonfire Seek and Destroy page

livike:
- Added Solution 2, 
- enlarged hints for solution 2

### DIFF
--- a/Bonfire-Seek-and-Destroy.md
+++ b/Bonfire-Seek-and-Destroy.md
@@ -4,18 +4,21 @@ This problem is a bit tricky because you have to familiarize yourself with Argum
 ## Hint: 1
 You need to work with `arguments` as if it was a regular array. The best way is to convert it into one.
 
+For solution #2 The `arguments` object is an array-like object and have some behavior identical with a regular array. You can use this to identify each argument separately.
+
 ## Hint: 2
 You need to filter, this also means you need to create a callback function, one that checks if the element is on the `indexOf()`
 
 ## Hint: 3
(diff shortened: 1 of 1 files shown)


Rafase282:
I'm not so sure about the solution. If anything the only thing I can see from there teaching something new would be the fact that you can use a function that was already declared with filter. That's not new to me and it was common sense when I learned about filter but I'm not sure if others know this too.

I would like some feedback from @FreeCodeCamp/teamwiki @FreeCodeCamp/wiki-editor 

I would like to limit the code solutions to things that improve on the previous teach something more advance and have a base code with the bare minimum based on the links mentioned on the bonfires and such.

SaintPeter:
This is really not a very good solution.  I think it's harder to read and understand that the prior solution and while it is vaguely different, it's also less efficient.

livike:
Thanks for the feedbacks.
@Rafase282 : The callback function was created exclusively for the filter. I learned about filter() from the link mentioned on the bonfires and the first example was the callback function defined separately.
However I tried to put the function directly in filter, but the FCC code editor put the warning to not make functions within a loop. For this is the callback function separately.

@SaintPeter : personally, at first I find a little hard to understand the solution in the wiki, for this reason wanted to create my own, maybe less elegant and efficient.
I like the fact that in many wiki hint pages you can find many different ways to solve the problem. Looked like a good possibility to contribute and submit one different.
However if you find it so bad, we can not include it.

Rafase282:
@livike the thing is with filter, you don't need a loop. It has one built inside which is why you don't see on one the example I provided. They are redundant and using them defeats the purpose of using filter at all, at least for this problem.

PR #274, opened 2016-01-16
Added compact solution for the challenge.

sgalizia:
The added solution is a compact version, similar to solution 3, that
uses the same logic, but uses the compact version of the if statement to
reduce the size of the function.


sgalizia:
I am going to close this pull request and reopen with the updated layout for the page. It won't let me create a new pull request otherwise.

hallaathrad:
And now, as pièce de résistance, let's squash the multiple commit out of this PR :100: 
Check [this one out](https://github.com/FreeCodeCamp/FreeCodeCamp/wiki/git-rebase#squashing-multiple-commits-into-one). Let me know if you need any help.

PR #281, opened 2016-01-19
Add cities, Jan 18-20 2016

hallaathrad:


hallaathrad:
How do I create a new label, anyway... I don't think enhancement applies as much as "content update" or something similar would

hallaathrad:
Please merge for me, someone? :(

PR #285, opened 2016-01-21
Bonfire-Caesars-Cipher --- New solution with map() method

aemb:
I'm not sure about my english skills. Explanations section may need a lookup :)

### DIFF
--- a/Bonfire-Caesars-Cipher.md
+++ b/Bonfire-Caesars-Cipher.md
@@ -33,12 +33,12 @@ function rot13(str) {
             nstr += String.fromCharCode(str.charCodeAt(i));
             continue;
         }
-        //If the ASCII gets out of range add 26 to it. (ASCII+26-13)
+        //N = ASCII 78, if the character code is less than 78, shift forward 13 places
         if(str.charCodeAt(i) < 78){
             nstr += String.fromCharCode(str.charCodeAt(i) + 13);
         }
         else{
-            //Shift the character 13 places backward
+            //Otherwise shift the character 13 places backward
             nstr += String.fromCharCode(str.charCodeAt(i) - 13);
 
         }
@@ -48,9 +48,33 @@ function rot13(str) {
 }
 ```
 
+#### Second solution (advanced)
+
+```js
+function rot13(str) {
+  // Split str into a character array
+  return str.split('')
+  // Iterate over each character in the array
+    .map.call(str, function(char) {
+      // Convert char to a character code
+      x = char.charCodeAt(0);
+      // Checks if character lies between A-Z
+      if (x < 65 || x > 91) {
+        return String.fromCharCode(x);  // Return un-converted character
+      }
+      //N = ASCII 78, if the character code is less than 78, shift forward 13 places
+      else if (x < 78) {
+        return String.fromCharCode(x + 13);
+      }
+      // Otherwise shift the character 13 places backward
+      return String.fromCharCode(x - 13);
+    }).join('');  // Rejoin the array into a string
+}
+```
+
 # Code Explanation:
 #### First solution
-- A string variable nstr is declared and initialized to store the
+- A string variable `nstr` is declared and initialized to store the
 decoded string.
 - The for loop is used to loop through each character of the input string.
 - If the character is not uppercase English alphabets(i.e. its ascii doesn't lie between 65 and 91 ), we'll leave it
@@ -63,6 +87,6 @@ range when subtracted by 13 so we'll add 26(number of letters in
 e.g. M(77) ↔ 77-13 = 64(Not an English alphabet) +26 = 90 ↔ Z(90)
 
 # Credits:
-If you found this page useful, you can give thanks by copying and pasting this on the main chat:  **`thanks @anuragaryan for your help with Bonfire: Caesar's Cipher`**
+If you found this page useful, you can give thanks by copying and pasting this on the main chat:  **`thanks @anuragaryan and @SaintPeter for your help with Bonfire: Caesar's Cipher`**
 
 > **NOTE:** Please add your username only if you have added any **relevant main contents** to the wiki page. (Please don't remove any existing usernames.)


Rafase282:
please, [squash](https://github.com/FreeCodeCamp/FreeCodeCamp/wiki/git-rebase#squashing-multiple-commits-into-one) your commits before force pushing.

Bonfire-Caesars-Cipher.md
```diff
@@ -48,6 +48,41 @@ function rot13(str) {
...
+    var strArray = [].map.call(str, function(x) {
+        return x.charCodeAt(0); 
+    });
+  
+    // Take the previous array (strArray) and then shift all carracters codes 
+    // and place them in a string variable
+    map.call(strArray, function(x) {
+        //Checks if character lies between A-Z
+        if (x < 65 || x > 91) {
+            newStr += String.fromCharCode(x);
```
SaintPeter: This is a bad an incorrect use of map.

If you're modifying data outside of the map, you should be using `.forEach`.

SaintPeter:
Here is a proper use of map:

```js
function rot13(str) {
    // Split str into a character array
  return str.split('')
       // Iterate over each character in the array
      .map.call(strArray, function(char) {
        // Convert char to a character code
        x = char.charCodeAt(0);
        // Checks if character lies between A-Z
        if (x < 65 || x > 91) {
            return String.fromCharCode(x);  // Return un-converted character
        }
        // If the ASCII gets out of range add 26 to it. (ASCII+26-13)
        else if (x < 78) {
            return String.fromCharCode(x + 13);
        }
        // Otherwise Shift the character 13 places backward
          return String.fromCharCode(x - 13);
        }
    }).join('');  // Rejoin the array into a string
}
```

aemb:
Thanks @SaintPeter. That one is much cleaner :+1:

aemb:
@Rafase282 I'm not sure that I have done it correctly. Let me know if there is a way to know if I did it the right way.

Bonfire-Caesars-Cipher.md
```diff
@@ -62,7 +86,15 @@ range when subtracted by 13 so we'll add 26(number of letters in
         English alphabets) to it so that after A it'll go back to Z.
 e.g. M(77) ↔ 77-13 = 64(Not an English alphabet) +26 = 90 ↔ Z(90)
 
+#### Second solution
+- Using `String.prototype.split()` will help us separate the string `str` into substrings. This method splits a string object into an array.
+- We are using the `Array.prototype.map()` because it calls a provided callback function once for each element in an array, in order, and
```
SaintPeter: None of this explanation looks right anymore

Bonfire-Caesars-Cipher.md
```diff
@@ -62,7 +86,15 @@ range when subtracted by 13 so we'll add 26(number of letters in
...
+ constructs a new array from the results. In this case `char` equal each element that `map()` gets from the array.
+- We are also using `Function.prototype.call()` to call `map()` on the argument `str`.
```
SaintPeter: This is not happening anymore

Bonfire-Caesars-Cipher.md
SaintPeter: We're not really . . . well, kinda.

SaintPeter:
It might be best to just leave the inline comments as the explanation for the second solution.

aemb:
It looks like commits are squashed the right way now.
I removed the section @SaintPeter. Thanks for your help. I feel like I will take some time before feeling at ease with map(). I just started using it and I'm on Basic Algorithm Scripting (50 hours). Maybe is to early for me.

hallaathrad:
Nope... the earlier you can use map instead of regular loops, the best. It's only a matter of getting it right, and though it'll take a while until you get the hang of it, it's absolutely worth it.
[Here's why](http://www.sitepoint.com/quick-tip-stop-writing-loops-start-thinking-with-maps/)

Bonfire-Caesars-Cipher.md
```diff
@@ -48,9 +48,33 @@ function rot13(str) {
...
+      // Otherwise Shift the character 13 places backward
+      return String.fromCharCode(x - 13);
+    }).join('');  // Rejoin the array into a string
+}
+```
+
 # Code Explanation:
 #### First solution
-- A string variable nstr is declared and initialized to store the
+- A string variable `nstr` is declared and initialized to store the
```
Rafase282: That is not the case with the current code.

Bonfire-Caesars-Cipher.md
```diff
@@ -48,9 +48,33 @@ function rot13(str) {
...
+  return str.split('')
+  // Iterate over each character in the array
+    .map.call(str, function(char) {
+      // Convert char to a character code
+      x = char.charCodeAt(0);
+      // Checks if character lies between A-Z
+      if (x < 65 || x > 91) {
+        return String.fromCharCode(x);  // Return un-converted character
+      }
+      // If the ASCII gets out of range add 26 to it. (ASCII+26-13)
```
SaintPeter: This is confusing - you say "add 26 to it", but it's really adding 13?   
Maybe something like:
> N = ASCII 78, if the letter is less than 78, shift forward 13 places

Or something like that?

SaintPeter:
Once we get the comment cleared up, I'm good with the second solution.  I ran it against the site just to make sure it passed.  :wink:

Rafase282:
By the way, my code:

```js
  return str.toUpperCase().split('').map(function(subStr) {
    return decode(subStr.charCodeAt(0));
  }).join('');

  function decode(arg) {
    // Takes care of characters that are not in [A-Z] such as ! and ? and decodes [A-Z]
    var decoded = 0;
    if (arg < 65 || arg > 90) {
      return String.fromCharCode(arg);
    } else {
      decoded = (arg + 13) % 91;
    }
    return String.fromCharCode(decoded < 65 ? decoded += 65 : decoded);
  }

}
rot13("SERR CVMMN!");
```

aemb:
@Rafase282 This would be nice with an explanation on the following lines for newbies like me :)
The assignment on `decoded` value looks like black magic xD
```{javascript}
decoded = (arg + 13) % 91;
//add
return String.fromCharCode(decoded < 65 ? decoded += 65 : decoded);
```

Thanks @hallaathrad for the article.

Rafase282:
@aemb lol thanks. I was not intending on adding my solution as a similar one was there but if you and others think it will add value then  I can add it.

The first line  basically adds 13 and then does the mod with 91 so it will go on the right range, how did I get the right value? With lots of tries, had my math been better I could probably had reversed engineered a math formula in a more proficient way.

So if the number is less than 65 then it means it is either 0 or a greater number which should have gone from 65 so i just add 65 to it and it will fall on the right place. Basically is a way to make it circular. Something I learned in me Freshman year of college with python and never used until now.

If you need more details or numbers let me know or add logs to the code.

Let me know if you want to merge now or if you want to integrate my code.

aemb:
What I am curious about is how getting things done in that way does impact on your code. Does that improve the execution of the code and its organisation? @Rafase282 I think they may be benefits to do it like this. The first one being learning something you didn't knew before. I would like to integrate your code but it will need some really precise explanations and well written to be affordable to everyone (TODO --> working on).

I also just realized you have done something I wouldn't have done in javascript : you wrote a function `decode` after calling it in the return statement and it works (I started with ProcessingJS sinces I have always thought that functions had to be written before being called).
I am learning more magic things with this problem :)

SaintPeter:
@aemb I do think that it improves the organization of the code, although maybe not the efficiency.  An important consideration when writing code is to make it "self documenting" and maintainable.  You will only write it once, but many others will read it.

RE: The order of definitions
Javascript has a neat feature called [Hoisting](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/var#var_hoisting).  JS does two passes on the code:  In the first it collects all variable definitions and "hoists" them to the top of the scope where they are defined.  This can lead to some really odd looking, but valid code:
```js
a = 5;
b = 7;
console.log(a,b); // 6 7
var a;
var b;
```

More here: http://www.sitepoint.com/demystifying-javascript-variable-scope-hoisting/

aemb:
Thanks @SaintPeter
I needed to know this :)

PR #300, opened 2016-01-27
Add a guide to use Browsec plugin for Chrome

raisedadead:
This article is a guide to resolve issues related to the "man in the middle" attacks,
faced by some campers.

The article aims to help users to install the Browsec plugin for Chrome.
Please refer FreeCodeCamp/FreeCodeCamp#5999 for details.


raisedadead:
@SaintPeter @BerkeleyTrue  @AkiraLaine 

Guys can you check this? Please let me know if I need to make any changes.
Thanks.

Using-browsec-for-securing-your-connection-to-fcc.md
```diff
@@ -0,0 +1,63 @@
...
+
+Well for some campers this is a classic case of man in the middle attack, for instance in the issue \#5999 the ISP some of the campers from India are using is deliberately, injecting Ads into the user's visited websites that's causing the issue.
+
+Sometimes, some ISPs use a similar strategy to cache the content of the websites visited by the consumers, so that they can serve the same to other consumers, this saves them bandwidth costs, while making the websites available faster.
+
+But when this is not done correctly (or is done malaciously as in the first case), this blocks essential content for websites to run seemlessly.
+
+## What's the fix, then?
+
+Simple, we need to anonymize ourselves i.e connect to the FCC website in a secure way, by-passing the ISP, or the firewall caches, whatsoever.
```
SaintPeter: Isn't it more accurate to say that we need to encrypt our traffic in a way that it cannot be modified during transmission?

Using-browsec-for-securing-your-connection-to-fcc.md
```diff
@@ -0,0 +1,63 @@
...
+Lately some of the campers started facing a strange problem, namely content blocking by some Internet Service Providers (ISP), leading to the [FreeCodeCamp.com](http://freecodecamp.com) (FCC) website being broken sometimes.
+
+This came soon to light with some of the issues logged by campers :
+- [\#5999 : Indian ISP's injecting Ads breaks site](https://github.com/FreeCodeCamp/FreeCodeCamp/issues/5999)
+- [\#6122 : Serious Problem using the FCC IDE, console showing error...](https://github.com/FreeCodeCamp/FreeCodeCamp/issues/6122)
+- [\#6381 : Unable to view the Code editor in the browser](https://github.com/FreeCodeCamp/FreeCodeCamp/issues/6381)
+
+## TL;DR : What's Wrong, again?
+
+Well for some campers this is a classic case of man in the middle attack, for instance in the issue \#5999 the ISP some of the campers from India are using is deliberately, injecting Ads into the user's visited websites that's causing the issue.
```
SaintPeter: Add Wikipedia Link on `man in the middle attack`
https://en.wikipedia.org/wiki/Man-in-the-middle_attack

You may want to explain that they are modifying websites as they are transmitted.  In some cases this is to serve ads, and in other cases . . .  etc.

SaintPeter: Grammar, add commas for an aside:
> Well, for some campers, this is 

Grammar, remove comma:
> using is deliberately injecting Ads into . . .

Using-browsec-for-securing-your-connection-to-fcc.md
```diff
@@ -0,0 +1,63 @@
+# Using Browsec for Securing your Connection to FreeCodeCamp.com
+
+## Why do I need an plugin?
+Lately some of the campers started facing a strange problem, namely content blocking by some Internet Service Providers (ISP), leading to the [FreeCodeCamp.com](http://freecodecamp.com) (FCC) website being broken sometimes.
```
SaintPeter: Grammar:
> facing strange problems.

In general we're not so much seeing "content blocking" as "content modification".

but the real "problem" is not so much that the content is being modified as that the modification is breaking the site.  It's a fine distinction, but important.

SaintPeter:
Aside from comments above, this looks pretty solid.

One thing you may want to do is explain how Browser Sec works - what it is that it's doing.  I presume that it's using a 3rd party service to provide VPN.  You should explain what that means - that they're sending all their data through a 3rd party and that it may slow things down, etc.  It's not without tradeoffs.

raisedadead:
Thanks for the review @SaintPeter. I'll do the changes. 

Also I found one more change that I need to add, really sorry having missed that in the first place, which is about a **Portable Firefox bundled with an Browsec Extension**. That covers one more browser. I'll test its use if I get time, but mostly should work.

Just one more query though, about adding how Browsec works (explaining the details with caveats), unfortunately I am not sure, how it actually does it behind the curtains, its closed source and not much is available on its official web site, so for the sake of this article lets put it this way, so that it covers the below from its feature set, along with things like you mentioned (speed for instance, or possible tracking by browsec's servers on non secure websites, etc.)

how

What do you say?

Thanks again.

SaintPeter:
Well, I think that you need to at least mention in general what we THINK is going on - they're talking about using a VPN, I think.  I don't know if we need to explicitly disclaim that we are not providing or endorsing the software.  We're just suggesting it might work.

Go ahead and make the changes you think are appropriate and be sure you comment after you amend your commit.

AkiraLaine:
It looks good, but don't complicate things too much. Just say "this is the problem and here is the solution".

raisedadead:
@SaintPeter Updated with the changes, the [most recent one is here](https://github.com/raisedadead/wiki/blob/wikiguide/using-browsec-for-securing-your-connection-to-fcc/Using-browsec-for-securing-your-connection-to-fcc.md)

Changes done:
1. Modified the text accommodating your review comments.
2. Added the Firefox installation.
3. Added a disclaimer (You are right, we should do this.) 
4. Did some pretty printing to the markdown too (the images were too large and the text was a bit off).

Let me know how you feel about it.

@AkiraLaine : I agree, we should be straight, but while the wiki should be as simple as possible, yet it should cover, some obvious queries that might come into the reader's mind, like "How it works?", "What maybe the caveats?". However, I am really open to any changes that you propose. 
Please let me know. :smiley: 

Thanks again both of you.

Using-browsec-for-securing-your-connection-to-fcc.md
```diff
@@ -0,0 +1,89 @@
...
+
+## TL;DR : What's Wrong, again?
+
+Well, for some campers, this is a classic case of [man in the middle attack](https://en.wikipedia.org/wiki/Man-in-the-middle_attack). 
+
+For instance, as seen in the issue \#5999, the ISP of some campers from India is deliberately injecting Ads into the user's visited websites that's causing problems.
+
+While, in other cases, ISPs sometimes use a similar strategy to cache the websites visited by consumers, so that they can serve the cached content to other consumers visiting the same websites, this saves them bandwidth costs, while making websites available faster.
+
+But when this is not done correctly (or is done malaciously as in the first case), this modifies essential content for websites, and prevents them to run seemlessly.
```
SaintPeter: Grammar: 
> and prevents them from running properly.

Using-browsec-for-securing-your-connection-to-fcc.md
```diff
@@ -0,0 +1,89 @@
...
+Of course yes. Why not? However remember that this does not make you invisible to law, so be careful of what you are doing! :wink:
+
+### What if this doesn't work for me?
+Please let us know in the [Help Chat](https://gitter.im/FreeCodeCamp/Help), we will try our best to find a workaround.
+
+---
+
+##### *Disclaimer*
+
+***FreeCodeCamp does NOT endorse any of the softwares mentioned in this article, please use them at your own discretion. Some VPN clients may track your activity, also there could be side-effects such as slow speeds or delay in page loads e.t.c.***
```
SaintPeter: It would be "etc", but I suggest just remove `e.t.c.` entirely.  Just put a period and be done with it.
> such as slow speeds or delays in page loads.

Using-browsec-for-securing-your-connection-to-fcc.md
```diff
@@ -0,0 +1,89 @@
...
+
+For instance, as seen in the issue \#5999, the ISP of some campers from India is deliberately injecting Ads into the user's visited websites that's causing problems.
+
+While, in other cases, ISPs sometimes use a similar strategy to cache the websites visited by consumers, so that they can serve the cached content to other consumers visiting the same websites, this saves them bandwidth costs, while making websites available faster.
+
+But when this is not done correctly (or is done malaciously as in the first case), this modifies essential content for websites, and prevents them to run seemlessly.
+
+## What's the fix, then?
+
+Simple, we need to anonymize ourselves i.e connect to the FCC website, by encrypting our traffic so it cannot be modified during transmission. We need to by-pass the ISP, or the firewall caches, whatsoever.
```
SaintPeter: "whatsoever" is not used properly here.  

How about this alternate sentence:
> By encrypting our traffic we bypass the ISP's ability to modify or cache our content as it passes through their infrastructure.

SaintPeter: Although that makes it redundant.  Why do we need this second sentence at all?

I'm not sure that we're really "anonymizing" our traffic as much as encrypting it so it cannot be intercepted and modified.

SaintPeter:
Sorry for the "death by 1000 cuts" method of editing.  I think with these few minor changes we're good to go.

This article looks really good and I appreciate the time you've put into writing it.

raisedadead:
@SaintPeter : Done with the changes.

I don't mind any no of suggestions, the fact that we are doing this over and over again to such small details, is itself the proof of the quality of the article or code that we want to be out there.

Thanks again for the comments. :smiley:

SaintPeter:
Fab!  You're the man!

PR #312, opened 2016-02-01
added a solution

UniverSoulSun:
Its a bit similar to the recursive solution without it being recursive and a bit simpler.
I've tested it up to 5000 and it probably could go higher.
Not sure it can get that prize money for finding a new prime number the way the FCC editor was warning me of a infinite loop when I tested it with 10000 lol!


Rafase282:
There were conflicts so I decided to close it.

PR #319, opened 2016-02-05
Add declarative solution

hallaathrad:


Checkpoint-Profile-Lookup.md
```diff
@@ -59,7 +60,24 @@ for (var x = 0; x < contacts.length; x++){
 return "No such contact";
 ```
 
+#### Second solution
+(Declarative programming)
+
+```
+  var result = "No such contact";
+  contacts.map(function (contact) {
```
SaintPeter: When you are not modifying the contents of the calling array (`contacts`), you should use `forEach`.

I don't really like this solution because it forces you to loop through all of the contacts even if you find the contact.  Imagine that you had 10000 contacts and the answer was in the first one?  What a waste that would be.  This is one of the rare cases where functional programming doesn't make sense.

You could use `.some` instead:
```
contacts.some(funciton(contact) {
    if (contact.firstName == firstName) {
      if (contact[prop]){
        result = contact[prop];
      }
     else result = "No such property";
    return true;  // Exit the loop
    }
   return false; // Keep going
});
return result;
}

File: Checkpoint-Profile-Lookup.md [hunk 2]
```diff
@@ -74,9 +92,12 @@ return "No such contact";
 - `"likes"` is found within the first object, so the second `if` statement returns true
 - The value of `"likes"` is returned - `"Pizza", "Coding", "Brownie Points"`
 
-
+####Second solution
+(Declarative programming)
+Pretty much the same done in the previous solution, except without telling the machine how to do its job.
```
SaintPeter: Pretty dismissive.  If it's basically the same thing, why do we even care?  Why is this solution better?
>  except without telling the machine how to do its job.

This is not meaningful commentary.   Explain what is going on and why we'd do this or don't add the solution at all.

PR #332, opened 2016-02-09
Replace link for Algorithm intro

alistermada:
The Meet Bonfire ("Meet Algorithm") coding challenge has been renamed and replaced with slides so I deleted the solution for it and replaced it with a link to the new challenge: [Get Set for our Algorithm Challenges](http://www.freecodecamp.com/challenges/get-set-for-our-algorithm-challenges).


Rafase282:
I think it is okay to keep the page, extra information on a screenshot never hurt nobody.

alistermada:
@Rafase282 Kinda misleading to have a link on the wiki to a challenge called "Meet Algorithm" when it doesn't exist in the curriculum. I think it should at least still be renamed to "Get Set for our Algorithm Challenges" to match the new FCC map.

Rafase282:
If you look at the screenshot it is the same. At most replacing the link to point there but the original point was to add extra help so having the actual page remain is good.

alistermada:
My intention to replace this was for the following reasons:

1. Since the recent major FCC UI update, many of the challenges have been renamed, including "Meet Bonfires" which is now "Get Set for our Algorithm Challenges".
image
2. "Meet Bonfires" was a coding challenge, so the information on the page made sense. However, the new challenge is now a slide challenge so there is no code input from the user. This would make the solution irrelevant and it doesn't help anybody.
3. Challenges under the [Gear Up for Success](https://github.com/FreeCodeCamp/FreeCodeCamp/wiki/Map#user-content-gear-up-for-success) section of the map page are all slide challenges as well and they all link to their original slides. I was just following this formula for consistency.
4. Yes, there is a screenshot of the old "Meet Bonfire" in the slide challenge, but I believe that must have been oversight or just used an example. As explained before, the coding challenge does not exist anymore so there should not a solution for it.

Even if the "solution" were to be kept, it's still the wrong name. Please at least allow the name "Meet Algorithms" to be changed to "Get Set for our Algorithm Challenges".

Rafase282:
The others used links because of the time when they were created. You could link

Rafase282:
If the link is changed and the file renamed then it will be rendered obsolete since it will be harder to fine but given that it didn't have much explanation to beging with is fine.

As for the others, links were used because no pages were created before and they were self explanatory for the most part so links were used.

PR #333, opened 2016-02-09
Fix broken links on map

alistermada:
Updates challenge links on the map.


hallaathrad:
:+1: w00t

Rafase282:
Blounder on my part, the links seem correct now, I only tested a few but by format they are right.

PR #339, opened 2016-02-12
Small typo

hallaathrad:
Also, where's the calculator challenge? :(


Rafase282:
What about it? Was it removed or they were going to add a new one?

PR #342, opened 2016-02-12
updated fb header generic image

hallaathrad:
Because we don't need jurassic park anymore...


Rafase282:
I like the Jurassic Park theme personally but I'll merge.

PR #345, opened 2016-02-13
Update Algorithm-Convert-HTML-Entities.md

ctsstc:
I added a more more dynamic solution that allows for only needing to change the dictionary.

### DIFF
--- a/Algorithm-Convert-HTML-Entities.md
+++ b/Algorithm-Convert-HTML-Entities.md
@@ -71,6 +71,28 @@ function convert(str) {
   return str;
 }
 ```
+
+##Another Solution
+```js
+function convert(str) {
+    
+    //map of key:value pairs
+    var dictionary = {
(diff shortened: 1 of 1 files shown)


Rafase282:
The second and third solution seems to be similar without checking into details. Could you please try to merge both approach to get the best of both similar solutions?

Also please understand that when I uploaded the first solution and made the page there was nothing before and it was straight from my own wiki where I used to explain the code on comments. For this wiki, we need to provide more details beyond the code comments to explain what is being done and what you are aiming to teach with your solution. 

So I'll ask you to make those changes.

1. Merge solution 2 & 3
2. Explain the code
3. Reorganize the page to keep it clean, like Solution # and so on instead of "Another Solution"
4. The code explanation would be on the section where it would be not just bellow each solution. This might change in the future, it is something to discuss on the wiki room.

PR #369, opened 2016-02-18
Fix filename and title

abhisekp:
Rename title **Updates February 11 2015** to **Updates February 11 2016**


abhisekp:
@Rafase282 woah! That was fast :zap:

Rafase282:
@abhisekp I'm having trouble staying asleep through the night so I check my email when I wakeup.

Issue #344, opened 2016-02-13
Need a JSON or CSV file for List of FCC cities


samosale:
https://github.com/FreeCodeCamp/freecodecamp/wiki/List-of-Free-Code-Camp-city-based-Campsites

If you go to that page you will find the list of all cities. I need that data in csv or json. If you go to google maps you will see that it is using my.csv to generate the chart. It would be perfect if I had that filewith or without coordinates. tnx

Rafase282:
Hello, didn't you already do this? is it okay to close this issue now?

PR #389, opened 2016-02-26
#344 campsites to json - lat/lon blank

LibertysYarn:


Rafase282:
@hallaathrad can you check this out?

hallaathrad:
Looks legit to me, though I wonder about sustainability, considering we add so many cities every single day. @Rafase282 

@LibertysYarn how does this json handle the `"lat": , "lon"; `? do these come off as undefined? or blanks? or does it simply turn into an error?

LibertysYarn:
I'm pondering on it. Also trying to estimate length of time for me to fill
all of those in. I can change them all to null for now.

Rafase282:
What's the status on this? Ready to Squash and merge or there is no resolution yet?

PR #414, opened 2016-02-28
Git-pull-vs-git-fetch.md

aj07:
This file describes about he difference between git pull and git fetch.

### DIFF
--- a/Git-Pull-vs-Git-Fetch.md
+++ b/Git-Pull-vs-Git-Fetch.md
@@ -1,0 +1,18 @@
+<h1>Git Pull vs Git Fetch:</h1>
+
+These two commands are regularly used by git users. Let's see the difference between both commands.
+
+
+A clone is simply a copy of a repository. It is something like where you can download source code from some other repository.
+<b>git fetch</b> is the command that says "bring my local copy of the remote repository up to date."
+<b>git pull</b> says "bring the changes in the remote repository where I keep my own code."
+Normally when use <b>git pull</b> , it does the same thing done by <b>git fetch</b> to bring the local copy of the remote repository(up-to-date), and then merging the changes into own code repo.Example
+ 
+     git pull origin ankur bugfix         
+
+
+The take away is to keep in mind that there are often at least three copies of a project on your workstation. One copy is your own repository with your own commit history. The second copy is your working copy where you are editing and building. The third copy is your local "cached" copy of a remote repository.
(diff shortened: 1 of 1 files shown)


Rafase282:
@aj07 the information is good but we prefer Markdown for Wiki articles. Let me know if you need help with formatting.

Git-Pull-vs-Git-Fetch.md
```diff
@@ -0,0 +1,18 @@
+<h1>Git Pull vs Git Fetch:</h1>
+
+These two commands are regularly used by git users. Let's see the difference between both commands.
+
+
+A clone is simply a copy of a repository. It is something like where you can download source code from some other repository.
```
batjko: I think the phrasing could be improved a little here.
"_It is something like where you can download source code from some other repository_." could be changed to something similar to this:
"It is similar to downloading source code from a different source that is not version-controlled."

Git-Pull-vs-Git-Fetch.md
```diff
@@ -0,0 +1,18 @@
...
+<b>git fetch</b> is the command that says "bring my local copy of the remote repository up to date."
```
batjko: I would be a bit more accurate here, because it does not really bring the local copy up-to-date. It only retrieves the latest information _about_ the remote repository.

How about something like this:
"Gets the latest meta-information, such as branches and commits, about the remote repository, without downloading the actual data."

Rafase282:
1. Please only use markdown. You can find a guide [here.](https://guides.github.com/features/mastering-markdown/)
2. Write the article as a guide, use third person instead of first person.
3. For the title, follow the guidelines found [here](https://github.com/FreeCodeCamp/FreeCodeCamp/wiki/PULL_REQUEST_TEMPLATE) (it was recently updated to make it easier for others to avoid the same issue.)  Basically use something like `Title-For-The-Article.md`

Once you have implemented the changes suggested let us know.

Rafase282:
I'm closing due to inactivity. Feel free to reopen when you have the time @aj07

Git-Pull-vs-Git-Fetch.md
hallaathrad: I likez! thanks (using it)

hallaathrad:
See https://github.com/FreeCodeCamp/wiki/pull/525 for further corrections.

Git-Pull-vs-Git-Fetch.md
aj07: added

PR #424, opened 2016-02-29
Add Wiki on How to Be Comfortable Using Reduce With simple examples

alayek:
I am adding an article on how to use Array.prototype.reduce() in day-to-day problems; and get comfortable with it.

### DIFF
--- a/Reduce-made-easy.md
+++ b/Reduce-made-easy.md
@@ -1,0 +1,376 @@
+# Using `Array.prototype.reduce()` to Reduce Conceptual Boilerplate for Problems on Arrays
+
+That's quite a mouthful! It could have just been titled as **Use `Array.prototype.reduce()` to Solve Array Problems Easily** or **`Array.prototype.reduce()` FTW!**. That would have been so much easier to read and parse.
+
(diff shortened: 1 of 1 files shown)


Reduce-made-easy.md
```diff
@@ -0,0 +1,370 @@
...
+
+```js
+var arr = [10, 20, 30, 60];
+arr.reduce(function(acc, item){
+   console.log(acc, item, this);
+   return acc;
+}, 0);
+```
+
+If you are in strict mode, it would return `undefined` as value of `this`. Otherwise, in-browser, it would point to `window` object as `this`.
```
raisedadead: Can we point to the MDN link in the text "strict mode"? 
It would be a good clarification for someone who isn't aware of it.

Apart from that this is an excellent write up. Just [squash](https://github.com/FreeCodeCamp/wiki/blob/master/git-rebase.md#squashing-multiple-commits-into-one) the commits and we are good to go!

Thanks for the article @alayek !

alayek:
Ok, done! Squashed 8 commits into 1.

raisedadead:
Supeeer! :+1:

PR #425, opened 2016-02-29
Better header for Code Formatting

abhisekp:
People are too lazy to read the whole thing. They just want to paste their code in pretty format ASAP.

- Moved the multi-line formatting method to top so that people don't accidentally use the single line formatting.
- Modified the heading so that it better shows the intended behavior.


raisedadead:
Thanks @abhisekp :+1:

PR #437, opened 2016-02-29
Update to Challenge Solution Template

abhisekp:
Major update to Challenge Solution Template

- Renamed `Checkpoint-Template.md` to `Challenge-Solution-Template.md`
- Markdown formatting according to relevance and compatible with @camperbot 

> Needs review and improvements before merging.


Rafase282:
I would suggest:

1. **Basic Code Solution**
2. **Intermediate Code Solution**
3. **Advanced Code Solution**

Instead of Code Solution #

abhisekp:
@Rafase282 There might be multiple code solutions for a given category. Maybe a label such the following would be nice.

## Code Solution 1: **`[basic]`**

```js
// write solution 1 code here
```

### Code Explanation:
> Explain solution 1 code here

## Code Solution 2: **`[basic]`**

```js
// write solution 2 code here
```

### Code Explanation:
> Explain solution 2 code here

## Code Solution 3: **`[advanced]`**

```js
// write solution 3 code here
```

### Code Explanation:
> Explain solution 3 code here

Rafase282:
I said it because we have been talking and we don't want more than three solution per challenge. That would be too much. If new solutions come up, changes are that they will be similar to previous ones and if the new one has something new then the two similar ones should me merged if possible, otherwise keep the code that is deemed better for teaching purposes.

abhisekp:
And also in the Code explanation part, I would like to amend it as

### Code Explanation:
> Explain solution 3 code here
> add any relevant links

#### Relevant Links
- :point_right: [Link Title 1](http://example.com)
- :point_right: [Link Title 2](http://example.com)
- :point_right: [Link Title 3](http://example.com)

PR #463, opened 2016-03-01
Update typo and linkify method calls for Array Reduce

koustuvsinha:
Updated typo and linkified method calls for Array Reduce Wiki

### DIFF
--- a/Reduce-made-easy.md
+++ b/Reduce-made-easy.md
@@ -1,13 +1,13 @@
 # Using `Array.prototype.reduce()` to Reduce Conceptual Boilerplate for Problems on Arrays
 
-That's quite a mouthful! It could have just been titled as **Use `Array.prototype.reduce()` to Solve Array Problems Easily** or **`Array.prototype.reduce()` FTW!**. That would have been so much easier to read and parse.
+That's quite a mouthful! It could have just been titled as **Use [`Array.prototype.reduce()`](https://github.com/FreeCodeCamp/FreeCodeCamp/wiki/js-Array-prototype-reduce) to Solve Array Problems Easily** or **`Array.prototype.reduce()` FTW!**. That would have been so much easier to read and parse.
 
 But It was not. Loops in JavaScript are just like that. They are not terse, they make you beat around the bush for a while. As the joke goes, two things are most difficult in computer science - [cache invalidation](https://en.wikipedia.org/wiki/Cache_invalidation), [naming things](https://www.quora.com/Why-is-naming-things-hard-in-computer-science-and-how-can-it-can-be-made-easier), and [off-by-one error](https://en.wikipedia.org/wiki/Off-by-one_error).
 
 And then there is the danger of writing [asynchonous code inside a for loop without using closure from IIFE](http://stackoverflow.com/questions/11488014/asynchronous-process-inside-a-javascript-for-loop).
 
 This article would start with a claim - that you can avoid using a for-loop or while-loop to solve any `Array` related problems.
-Instead, you can solve all of them using `Array.prototype.reduce()`. If you wish to read forward; do make sure you know about recursive functions, and some of the cool functional tools like `map()` or `filter()`.
+Instead, you can solve all of them using `Array.prototype.reduce()`. If you wish to read forward; do make sure you know about recursive functions, and some of the cool functional tools like [`Array.prototype.map()`](https://github.com/FreeCodeCamp/FreeCodeCamp/wiki/js-Array-prototype-map) or [`Array.prototype.filter()`](https://github.com/FreeCodeCamp/FreeCodeCamp/wiki/js-Array-prototype-filter).
 
 Grand claims require grand evidence. So let's demonstrate how we can get accustomed with using `reduce()`.
 
@@ -85,7 +85,7 @@ arr.reduce(function(acc, item){
 }, initialValue);
 ```
 
-Now, let's find out what would the value of these `acc` and `item` are. We have mentioned earlier that the `reduce` is a replacement for itertaive constructs.
+Now, let's find out what would the value of these `acc` and `item` are. We have mentioned earlier that the `reduce` is a replacement for iterative constructs.
 
 It stands to reason that `reduce` would take your custom callback function; and iterate over the Array on which `reduce` has been invoked.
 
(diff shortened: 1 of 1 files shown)


alayek:
Swag @koustuvsinha!

@raisedadead @Rafase282 @QuincyLarson kindly take a look.

Reduce-made-easy.md
```diff
@@ -374,3 +374,7 @@ If you omit this; for the first iteration `acc` would be the first item in the a
 But in some complex situations, it would be better to visualize and formulate the solution in terms of some base - some initialization. However, if you are more comfortable without it, to each his own!
 
 If you have any further questions or suggestions, come join our [gitter chatroom](https://gitter.im/FreeCodeCamp/FreeCodeCamp); and tell us how you `reduce`!
+
+## List of Contributors
```
Rafase282: Please remove this.

Reduce-made-easy.md
```diff
@@ -1,13 +1,13 @@
 # Using `Array.prototype.reduce()` to Reduce Conceptual Boilerplate for Problems on Arrays
 
-That's quite a mouthful! It could have just been titled as **Use `Array.prototype.reduce()` to Solve Array Problems Easily** or **`Array.prototype.reduce()` FTW!**. That would have been so much easier to read and parse.
+That's quite a mouthful! It could have just been titled as **Use [`Array.prototype.reduce()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce) to Solve Array Problems Easily** or **`Array.prototype.reduce()` FTW!**. That would have been so much easier to read and parse.
```
Rafase282: [[`Array.prototype.reduce()`]](https://github.com/FreeCodeCamp/FreeCodeCamp/wiki/js-Array-prototype-reduce)

Reduce-made-easy.md
```diff
@@ -1,13 +1,13 @@
...
 
 And then there is the danger of writing [asynchonous code inside a for loop without using closure from IIFE](http://stackoverflow.com/questions/11488014/asynchronous-process-inside-a-javascript-for-loop).
 
 This article would start with a claim - that you can avoid using a for-loop or while-loop to solve any `Array` related problems.
-Instead, you can solve all of them using `Array.prototype.reduce()`. If you wish to read forward; do make sure you know about recursive functions, and some of the cool functional tools like `map()` or `filter()`.
+Instead, you can solve all of them using `Array.prototype.reduce()`. If you wish to read forward; do make sure you know about recursive functions, and some of the cool functional tools like [`map()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) or [`filter()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter).
```
Rafase282: [`Array.prototype.map()`](https://github.com/FreeCodeCamp/FreeCodeCamp/wiki/js-Array-prototype-map)
[`Array.prototype.filter()`](https://github.com/FreeCodeCamp/FreeCodeCamp/wiki/js-Array-prototype-filter)

Reduce-made-easy.md
```diff
@@ -135,9 +135,9 @@ This time the output changes to:
 
 As you can see, the value of `acc` would remain constant throughout. And that is expected - we are not altering the value of `acc` anywhere in the custom callback. We return whatever `reduce` makes available at a given iteration.
 
-But we did realize something - the value of `acc` for current iteration, would be the `return` value from custom callback from previous iteration. And ultimately, when the iteration is over; the final value of `acc` would be returned by `reduce` call.
+But we did realize something - the value of `acc` for current iteration, would be the `return` value from custom callback from previous iteration. And ultimately, when the iteration is over, the final value of `acc` would be returned by `reduce` call.
 
-This leaves only one important part in our understanding - the value of execution context, or `this`!
+This leaves only one important part in our understanding - the value of execution context, or [`this`](http://javascriptissexy.com/understand-javascripts-this-with-clarity-and-master-it/)!
```
Rafase282: This could be a topic for a future new article.

Reduce-made-easy.md
```diff
@@ -149,8 +149,8 @@ arr.reduce(function(acc, item){
 }, 0);
 ```
 
-If you are in [strict mode](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode), it would return `undefined` as value of `this`. Otherwise, in-browser, it would point to `window` object as `this`.
-Can we override and set it on our own, using `bind`? Sure! just use `bind` with the callback:
+If you are in [strict mode](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode), it would return `undefined` as value of `this`. Otherwise, in-browser, it would point to [`window`](https://developer.mozilla.org/en-US/docs/Web/API/Window) object as `this`.
+Can we override and set it on our own, using [`bind`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_objects/Function/bind)? Sure! just use `bind` with the callback:
```
Rafase282: This could be a topic for a future new article.

Reduce-made-easy.md
```diff
@@ -202,7 +202,7 @@ var arr = [20, 50, 5, 60];
 arr.reduce(Math.max, 0);
 ```
 
-but this would not work and would return `NaN`. Here's the reason - `acc` and `item` are not the **only** arguments to the custom callback. When you call `Math.max()` tries to call it on non-numeric arguments, resulting in `NaN`.
+but this would not work and would return `NaN`. Here's the reason - `acc` and `item` are not the **only** arguments to the custom callback. When you call [`Math.max()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/max) tries to call it on non-numeric arguments, resulting in `NaN`.
```
Rafase282: [`Math.max()`](https://github.com/FreeCodeCamp/FreeCodeCamp/wiki/Math.max)

Reduce-made-easy.md
koustuvsinha: should i leave it like this as of now?

Rafase282:
I added the links for the articles we have, please use them with the full name as provided, I think it is better to have the full name, as for the ones that we don't have link, while I said they would be good future articles, I'm not implying you **should** create them.

koustuvsinha:
@Rafase282 updated with changes. please review :)

raisedadead:
@koustuvsinha Please squash the commits.

koustuvsinha:
done @raisedadead

Reduce-made-easy.md
Rafase282: Yes you can leave it as it is for now.

PR #466, opened 2016-03-01
Update Algorithm-Caesars-Cipher.md

anuragaryan:
There was a mistake in my code regarding the ASCII range, rectified it in both the solutions. 
Thanks to @mauney from gitter.im for pointing it out.

### DIFF
--- a/Algorithm-Caesars-Cipher.md
+++ b/Algorithm-Caesars-Cipher.md
@@ -29,7 +29,7 @@ function rot13(str) {
     var nstr="";
     for(var i=0; i<str.length; i++){
         //Checks if character lies between A-Z
-        if(str.charCodeAt(i) < 65 || str.charCodeAt(i) > 91) {
+        if(str.charCodeAt(i) < 65 || str.charCodeAt(i) > 90) {
             nstr += String.fromCharCode(str.charCodeAt(i));
             continue;
         }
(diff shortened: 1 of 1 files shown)


Rafase282:
@anuragaryan Would you mind explaining it to me then? I tested the code and my code works, I don't know if you based your solution off mine or why did you also use 91 but I had a good reason for it when I did.

Both code still work prior to these changes.

anuragaryan:
@Rafase282 A-Z is 65-90 in ASCII. As per initial condition, it checks for ASCII>91, thus consider ASCII 91 i.e.'[' as an English Alphabet and shift it 13 places backward. Which shouldn't happen.

Rafase282:
@anuragaryan sorry I got confused with the code, I didn't submit my answer to that one, mine is:

```js
  return str.toUpperCase().split('').map(function(subStr) {
    return decode(subStr.charCodeAt(0));
  }).join('');

  function decode(arg) {
    // Takes care of characters that are not in [A-Z] such as ! and ? and decodes [A-Z]
    var decoded = 0;
    if (arg < 65 || arg > 90) {
      return String.fromCharCode(arg);
    } else {
      decoded = (arg + 13) % 91;
    }
    return String.fromCharCode(decoded < 65 ? decoded += 65 : decoded);
  }

}
rot13("SERR CVMMN!");
```

It works and the 91 is needed because of the way I went about with 90 it would go and give me 0 instead of 1 and mess things up.

anuragaryan:
I happens. :+1:

Issue #367, opened 2016-02-17
New Content for the Wiki


Rafase282:
List is a work in progress.
Based off the most popular StackOverflow anwsers and my own choices so far.

# Git Related
- [ ] [Undo a commit and redo](http://stackoverflow.com/questions/927358/how-do-you-undo-the-last-commit/927386#927386)
- [ ] [Amending the most recent commit message](http://stackoverflow.com/questions/179123/edit-an-incorrect-commit-message-in-git/179147#179147)

# Backend Related
- [ ] API Authentication & Authorization

# Frontend Related
- [ ] [Page Redirect using Jquery](http://stackoverflow.com/questions/503093/how-can-i-make-a-page-redirect-using-jquery/506004#506004)

# Curriculum Specific

# Others

QuincyLarson:
@Rafase282 This is an excellent start. These are the sorts of things I myself have looked up a number of times, so it would be great to be able to read them in a wiki format (rather than a stack overflow format, which is less user-friendly)

hallaathrad:
Can I request we document how to setup a front-end dev environment?
How to set up PostCSS to concatenate JS, nest and prefix css etc; eslint, jshint, uglify, live reload, all these in a build system like gulp+bower.

It's been bugging me not really being able to find something decent like that, specifically thought for those who stick to the front.

Rafase282:
@hallaathrad yes, just add it there.

raisedadead:
#379 up for 
- [x]  Undo a commit and redo
- [x]  Delete a Git branch both locally and remotely
- [x]  Checkout Remote Branch
- [x]  Force Git to overwrite local files on pull

Rafase282:
Guys, to make it easier for all of us, if you sent a PR for any of the issues here, mark it here, the same if you are creating an article, let us know by marking it. The only thing is that if for whatever reason it is not merged and it is closed then it will need to be unchecked.

Also adding more articles from the list on the gist is needed and appreciated.

thomasvaeth:
I recently wrote a blog post on getting a basic front end dev environment set up. I can write something on the basics of Gulp today.

hallaathrad:
@thomasvaeth you'll become my hero for the 15 minutes after you post it! I've been trying to get away from needing CodeKit for the longest while.

Rafase282:
I'll look forward tot hat too.

Rafase282:
@hallaathrad the filename is Gulp-Basics.md

[...]

PR #467, opened 2016-03-01
Added a new wiki article -  Push a new local branch to a remote Git repository

rishipuri:
This PR adds a new wiki article w.r.t #367 : Git-Push-a-new-local-branch-to-a-remote-repository.md

Lemme know if any changes needed in the content.

Thanks.

### DIFF
--- a/Git-Push-Local-To-Remote-Repository.md
+++ b/Git-Push-Local-To-Remote-Repository.md
@@ -1,0 +1,15 @@
+# Push a new local branch to a remote Git repository and track it too
+
+1. Create a new branch:
+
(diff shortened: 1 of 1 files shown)


Rafase282:
@rishipuri You can use single backticks for inline code, it is prefered. Also the filename needs to be changed follow this format `The-Format-We-Use.md`

All articles will be required to use a filename like that so is best to start now. A shorter filename would make it easier too.

Something like git push local to remote branch or add the new keyword if there is a similar article, which if there is I would suggest merging then content instead.

rishipuri:
@Rafase282 Updated and squashed the commits.

PR #470, opened 2016-03-02
Rewrite article on JS scopes

awesomeaniruddh:
The previous article barely had any description and wasn't very helpful.
This is a more in-depth exploration of scopes, and there's a good chance that it's super confusing as well, so let me know if I'm not being clear enough.

Also, I changed the file name from `js-scopes.md` to `js-Scopes.ms` because that seemed to be the convention that the other files were following, though I am not sure if that's correct.

#367

### DIFF
--- a/js-Scopes.md
+++ b/js-Scopes.md
@@ -1,0 +1,95 @@
+# Scopes in JavaScript
+
+If you've been programming in JavaScript for a while, you've undoubtedly run into a concept known as `scope`. What is `scope`? Why should you take the time to learn it?
+
+In programmer speak, `scope` is the **current context of execution**. Confused? Let's take a look at the following piece of code:
+
+```js
+1.  var foo = 'Hi, I am foo!';
+2.  
+3.  var baz = function () {
+4.     var bar = 'Hi, I am bar too!';
+5.     console.log(foo);
+6.  }
+7.  
+8.  baz(); // Hi, I am foo!
+9.  console.log(bar); // ReferenceError...
+```
+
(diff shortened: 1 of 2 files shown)


js-Scopes.md
```diff
@@ -0,0 +1,93 @@
+If you've been programming in JavaScript for a while, you've undoubtedly run into a concept known as `scope`. What is `scope`? Why should you take the time to learn it?
```
raisedadead: Could you add an article heading as well?

raisedadead:
@awesomeaniruddh Nice write up, well I was in need of this at least, :smile:

js-Scopes.md
```diff
@@ -0,0 +1,95 @@
...
+9.  console.log(bar); // ReferenceError...
+```
+
+This is a simple example, but it does a good job of illustrating what is known as *Lexical scope*. JavaScript, and almost every other programming language has a *Lexical scope*. There is another kind of scope known as *Dynamic scope*, but we won't be discussing that.
+
+Now, the term *Lexical scope* sounds fancy, but as you will see it's really simple in principle. In a Lexical Scope, there are two kinds of scopes: the *global scope* and a *local scope*.
+
+Before you type the first line of code in your program, a *global scope* is created for you. This contains all the variables that you declare in your program **outside any functions**.
+
+In the example above, the variable `foo` is in the global scope of the program, while the variabe `bar` is declared inside a function and is therefore **in the local scope of that function**.
```
raisedadead: Typo `variabe` > `variable`

raisedadead:
:+1:

PR #472, opened 2016-03-02
Update chatroom lists with new back-end languages

alayek:
Updating the list of chatrooms to include new back-end languages.


SaintPeter:
:+1:

PR #473, opened 2016-03-02
Add python Operators documentation

koustuvsinha:
Converting python tutorial docs from https://github.com/alayek/python3notes from ipynb to markdown, by preserving formatting. Continuing work from [this PR](https://github.com/alayek/python3notes/pull/1)

### DIFF
--- a/Python-Operators.md
+++ b/Python-Operators.md
@@ -1,0 +1,210 @@
+# Operations and Operators
+
+## Truth Value Testing
+
+[Python Docs - Truth Value Testing](https://docs.python.org/3/library/stdtypes.html#truth-value-testing)
+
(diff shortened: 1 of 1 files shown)


alayek:
@koustuvsinha great job! Looks good to me.

@Rafase282 @raisedadead please have a look.

Rafase282:
Looks good, I guess given that it is a tested tutorial there are no typos or gramatical errors, I was watching out for those. The only thing is to change the filename to have all first character uppercase for each word and to use markdown links as in [ title ] ( url ) for the links that are showing the full url.

This helps besides looking better, and hiding clutter because we can then change those links to point to internal articles once they are create without any visual changes. No much yet but we want to keep readers on our domain as much as possible.

PR #487, opened 2016-03-02
Text proofing

hallaathrad:
Minor style changes and typos.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,11 +1,20 @@
 # FreeCodeCamp - Wiki Pull Request Template
+
+----
+
+PLEASE REMOVE THIS WHOLE TEMPLATE BEFORE YOU SUBMIT YOUR PULL REQUEST
+
+----
+
 Make sure your contribution addresses the points mentioned below:
 
 **Filenames:**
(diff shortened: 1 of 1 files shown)


Rafase282:
Looks good to me. Thanks!

Rafase282:
@hallaathrad could you please mention that only the first letter of every word must be capitalized for the filename?

SaintPeter:
This is refereed to as 'Title Case'.  Words such as "and" "an" and "the" are not capitalized in title case.

raisedadead:
@hallaathrad Please add a line like below:

> REMOVE THIS TEMPLATE TEXT FROM THE DESCRIPTION BEFORE YOU SUBMIT

Many contributors are just leaving this template in the PR description.

Rafase282:
@SaintPeter yes I know but I wanted to make it easier for regex usage to find filenames that needs to be changed. Also I think filenames should be short, they are to be used in links and people using the search function will use keywords so while I'm not going as far as to ask not to use articles on filenames, the less the better.

Overall the aim is a short yet self descriptive filename/title with the title having more flexibility.

Rafase282:
Thanks for the changes.

raisedadead:
:+1: lets merge this

hallaathrad:
Will squash in 20. Thanks for your patience

PR #494, opened 2016-03-03
Changed FCC Wiki links to relative links

Rafase282:
All links are now relative links and this should be the new standard for internal links. This might not work on the local wiki but it will work just fine on the live one.


Rafase282:
Merged for better testing.

alayek:
Thanks @Rafase282 I started testing, and LGTM!

PR #495, opened 2016-03-03
Update help chat rooms with links to new languages

alayek:
This is needed so that the `explain rooms` command can fetch links to new rooms as well.


alayek:
@Rafase282 @raisedadead please take a look.

Rafase282:
Looks good to me.

PR #506, opened 2016-03-04
Created git amend document.

jktravis:
Created document for #367.

- [X] Amending the most recent commit message


Rafase282:
As stated on chat, please use a shorter filename. Other than that it looks fine to me. I was thinking that this could be a goo alternative to squashing as we could just amend the previous commit instead of writing more.

PR #513, opened 2016-03-05
added a small fix to drop-it algorithm, solution was creating an infi…

danielcodes:


Rafase282:
@danielcodes could you tell me what the issue was again? You shoudl add that tot he description and not just the commit message as it got cutoff.

danielcodes:
Ah, my bad.

Ok, so if you run the code snippet provided in the wiki with a small print statement (in the exercise page), 

```javascript
while(!func(arr[0]) ){
  console.log("The current element is ", arr[0]);
  arr.shift();
}
```

and open up your console, you'll see this,

```
4729 The current element is  undefined (4729 iterations)
```

This happens because of this test case,

```javascript
drop([1, 2, 3, 4], function(n) {return n > 5;}) should return [].
```

all the elements return false for the function, so eventually the array empties out and you're left with

```javascript
arr = []

//the loop continues
arr[0] = undefined
func(undefined) = false
!false = true

//it then enters to shift an empty array, and the loop repeats.
```

Somehow, the loop stops but it is a bug nonetheless.
To make the loop stop, add a check if there is an element in the array in the first place with,

```javascript
arr[0] != undefined
```

Hope that clears things up, @Rafase282

Rafase282:
I ran the original code and it works fine.

Imgur

I see now what you are referring to.

danielcodes:
Thanks.. but why was the change reverted?

PR #526, opened 2016-03-06
Added Java-Basics file.

trishulpani:
Java Core concepts - initial commit.

### DIFF
--- a/Java-Basics.md
+++ b/Java-Basics.md
@@ -1,0 +1,489 @@
+# Java Basics
+
+## Core Concepts
+
+### The Java Virtual Machine (JVM) 
+
+Java belongs to a family of languages called **Compiled Languages** (https://en.wikipedia.org/wiki/Compiled_language). Any code written in such a language needs to be converted (compiled) to an intermediate form that can then be understood by the host platform (the OS/platform in which the code runs). 
+
+For Java, this intermediate form is called **Bytecode** which is then interpreted by a runtime called a Java Virtual Machine (JVM). Think of JVM (https://docs.oracle.com/javase/specs/jvms/se7/html/) as a piece of software that does the hard work of running your Java code. It takes care of memory allocation, thread management, garbage collection and so much more. Apart from Java, it also supports (read: able to run) code written in languages such as Groovy, Scala etc.
+
+In Java, code is written and saved as `.java` files. The compiler (javac) operates on the java files and generates the equivalent Bytecode (`.class`) files. The `java` command would now be able to execute the Bytecode stored in the `.class` files. More on this later. 
+
+The following sections describe some of the basic building blocks of coding in Java.
+
+### Variables
+
+Variables store values. 
+
+In [Java](https://github.com/FreeCodeCamp/FreeCodeCamp/wiki/Java), variables are [_strongly typed_](https://en.wikipedia.org/wiki/Strong_and_weak_typing#Definitions_of_.22strong.22_or_.22weak.22), which means you have to define the type for each variable whenever you declare it. Otherwise, the compiler will throw error at [compile time](https://en.wikipedia.org/wiki/Compile_time). Therefore, each variable has an associate data type of either :
+
+* Primitive Type : `int`, `short`, `char`, `long`, `boolean`, `byte`, `float`, `double`
+* Wrapper Type : `Integer`, `Short`, `Char`, `Long`, `Boolean`, `Byte`, `Float`, `Double`
+* Object Type: `String`, `StringBuilder`, `Calendar`, `ArrayList` etc.
+
+We made a distinction between **Wrapper Type** and general **Object Type** for a reason - wrapper types are closely linked with their primitive counterparts via [autoboxing and unboxing](https://docs.oracle.com/javase/tutorial/java/data/autoboxing.html); but more on that later.
+
+Typically you can declare variables using the following syntax :
+
+```java
+//Primitive Data Type
+
+int i = 10;
+
+// Object Data Type
+//initiates an Float object with value 1.0
+// variable myFloat now points to the object
+Float myFloat = new Float(1.0);
+```
+
+But there are more to these variables, [read about them here](#TODO).
+
+### Classes & Objects
+
+Classes are groups of variables and operations on them. A class can have variables, methods (or functions) and constructors (or methods which are used to initiate, more on that later!).
+
+Think of a `Class` as a blueprint for creating something concrete. A `Class` tells you the 'what' and 'how' an `object` of that Class will look like once `instantiated`. In essence, it defines `properties` (say color, engine capacity) and `behavior` (stop, speed up, change gears, honk etc.) for a Car in this case. 
+
+Objects are _instances_ of a class. All objects are instances of a certain class. Imagine a class being a "template", which every Object copies to. When you create an Object, basically it creates a new object on the blueprint of a class. Now lets look at this from a little piece of code :
+
+```java
+// Car class
+public class Car {
+  // car name
+  private String name;
+	// car mannufacturer name
+	private String manufacturerName;
+  // constructor
+	public Car(String name, String man) {
+		this.name = name;
+		this.manufacturerName = man;
+	}
+  // getter method
+	public String getName() {
+		return name;
+	}
+  // getter method
+	public String getManufacturerName() {
+		return manufacturerName;
+	}
+
+  //setter method
+  public void setName(String name){
+     this.name = name;
+  }	
+}
+
+Car modelS = new Car("Model S","Tesla");
+
+System.out.println("Full Car Name = " + modelS.getManufacturerName() + " " + modelS.getName());
+// prints Tesla Model S
+```
+
+So, `Car` is a class, which has the fields or properties name and manufacturerName. `modelS` is an object of `Car` class. So `modelS` also has the same properties and methods.
+
+### Access Modifiers
+
+Ok, so if `modelS` has all the same properties, I should be able to access `name` or `manufacturerName` right?
+
+```java
+System.out.println(modelS.name);
+```
+Results in :
+
+```asciidoc
+Main.java:13: error: name has private access in Car
+		System.out.println(modelS.name);
+		                         ^
+1 error
+```
+
+Why? Notice we had mentioned `private` before the variable `name` in class `Car`. This marks the variable as `private` to the class which means that the variable cannot be modified or accessed from another class, even if an object of the class is instantiated. So, unless you provide a public `setter` method or any other `public` method to access the variable, there is no way it's visible to the outside world.
+
+There are other kinds of access modifiers such as public, protected, default etc. [Read more about them here](#TODO).
+
+### Constructors
+
+What's the point then? I should be able to store data in it right?
+
+That's when we use either **getter** (e.g., getName()) / **setter** (e.g., setName()) methods or in this case constructors to initialize a class. Basically every Java Class has a constructor, which is the method which is called first when any object of the class is initialized. Think of it as a bit of starter code.
+
+When you write a class without any constructor, then Java assumes it has a default constructor :
+
+```java
+
+public class Car {
+	private String name;
+}
+
+Car modelS = new Car();
+```
+
+This initializing with no parameters is a way of calling the default constructor. You can also have a default constructor written yourself this way :
+
+```java
+public class Car {
+	private String name;
+
+	public Car() {
+		name = "Tesla";
+	}
+}
+```
+
+Then, when calling `new Car()`, the variable `name` will get auto-initialized to `"Tesla"`.
+
+Like what `Constructor`s can do? [Dive into it more here](#TODO).
+
+### Methods
+
+`getName()` and `getManufacturerName()` are two "Getter" methods we have used here. Notice, unlike JavaScript, we **have** to define the return type of any method we write, otherwise it will fail at compile time. If you do not want a method to return anything, use `void` return type.
+
+```java
+public class Car {
+	private String name;
+
+	public void changeName() {
+		name = "Tesla";
+	}
+}
+```
+As with any other language, methods (or functions, if you are here from JS world) are used often for their modularity and reusability. [Read more about them here](#TODO).
+
+## Inheritance
+
+So great you have successfully created a Car class. But, wait, aren't Tesla cars supposed to be electric variants? I want an Electric car class, but it also should have the properties of the original `Car` class.
+
+Solution : **Inheritance**. Java provides a neat way to "inherit" parent properties :
+
+```java
+public class Car {
+	
+	private String name;
+	private String manufacturerName;
+
+	public Car(String name, String man) {
+		this.name = name;
+		this.manufacturerName = man;
+	}
+    // Getter method
+	public String getName() {
+		return name;
+	}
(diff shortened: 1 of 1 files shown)


Java-Basics.md
```diff
@@ -0,0 +1,442 @@
+# Java Basics
+
+## Core Concepts
+
+### The Java Virtual Machine (JVM) 
+
+Java belongs to a family of languages called **Compiled Languages** (https://en.wikipedia.org/wiki/Compiled_language). Any code written in such a language needs to be converted (compiled) to an intermediate form that can then be understood by the host platform (the OS/platform in which the code runs). For Java, this intermediate form is called **Bytecode** which is then interpreted by a runtime called a Java Virtual Machine (JVM). Think of JVM (https://docs.oracle.com/javase/specs/jvms/se7/html/) as a piece of software that does the hard work of running your Java code. It takes care of memory allocation, thread management, garbage collection and so much more. Apart from Java, it also supports (read: able to run) code written in languages such as Groovy, Scala etc.
```
alayek: Begin a new paragraph from

> For Java, this intermediate form is called[...]

It gets much easier to read, once you break your text in smaller paragraphs. It's called [breaking the wall of text](http://tweakyourbiz.com/marketing/2012/08/10/avoiding-the-wall-of-text-when-creating-content/).

Java-Basics.md
```diff
@@ -0,0 +1,442 @@
...
+
+In Java, code is written and saved as `.java` files. The files are then run through a compiler (javac) to convert them to Bytecode `.class` files. Standalone Java programs are then run using the `java` command. More about this later.
```
alayek: > The files are then run through a compiler (javac)

Not exactly a sentence that should appear in a technical writing. A file is a collection of strings, stored in memory. 

The word "run" is associated with something that can be executed.

Phrase it better; with perhaps something like

> The Java compiler operates on the file, to generate equivalent Bytecode, stored in `.class` files.

alayek: > Standalone Java programs are then run using the `java` command.

Be more explicit here. _Standalone Java programs_ are not described in the previous lines.

You could just say 

> `java` command would now execute the Bytecode stored in the `.class` files.

Java-Basics.md
```diff
@@ -0,0 +1,442 @@
...
+
+Variables store values. 
+
+In [Java](https://github.com/FreeCodeCamp/FreeCodeCamp/wiki/Java), variables are [_strongly typed_](https://en.wikipedia.org/wiki/Strong_and_weak_typing#Definitions_of_.22strong.22_or_.22weak.22), which means you have to define the type for each variable whenever you declare it. Otherwise, the compiler will throw error at [compile time](https://en.wikipedia.org/wiki/Compile_time). Therefore, each variable has an associate data type of either :
+
+* Primitive Type : `int`, `short`, `char`, `long`, `boolean`, `byte`, `float`, `double`
+* Wrapper Type : `Integer`, `Short`, `Char`, `Long`, `Boolean`, `Byte`, `Float`, `Double`
+* Object Type: `String`, `StringBuilder`, `Calendar`, `ArrayList` etc.
+
+We made a distinction between **Wrapper Type** and general **Object Type** for a reason - wrapper types are closely linked with their primitive counterparts via [autoboxing and unboxing](https://docs.oracle.com/javase/tutorial/java/data/autoboxing.html).
```
alayek: > primitive counterparts via [autoboxing and unboxing](https://docs.oracle.com/javase/tutorial/java/data/autoboxing.html).

Change the line to 

> primitive counterparts via [autoboxing and unboxing](https://docs.oracle.com/javase/tutorial/java/data/autoboxing.html); but more on that later.

This would clearly signal a beginner that they don't need to understand autoboxing-unboxing right away.

Java-Basics.md
```diff
@@ -0,0 +1,442 @@
...
+Results in :
+
+```asciidoc
+Main.java:13: error: name has private access in Car
+		System.out.println(modelS.name);
+		                         ^
+1 error
+```
+
+Why? Notice we had mentioned `private` before the variable `name` in class `Car`. So since its private, we cant touch it from outside!
```
alayek: > So since its private, we cant touch it from outside!

Clearly mark boundaries. _outside_ here, in this context, means that you cannot access its value or alter its value from another class, even if you have instantiated the former class as an object there.

Also, no need for

> So since

Just using one of them is enough.

Java-Basics.md
```diff
@@ -0,0 +1,442 @@
...
+
+There are other kinds of access Modifiers such as public, protected, default etc. [Read more about them here](#TODO).
```
alayek: > There are other kinds of access Modifiers

_M_ need not be capitalized.

Java-Basics.md
```diff
@@ -0,0 +1,442 @@
...
+
+### Constructors
+
+What's the point then? I should be able to store data in it right?
+
+Thats when we use either Getter / Setter methods or in this case Constructors to initialize a class. Basically every Java Class has a constructor, which is the method which is called first when any object of the class is initialized. Think of it as a bit of starter code.
```
alayek: > Thats when we use either Getter / Setter methods 

It should be _That's_, short for _That is_

Also, no reason why _Getter_ and _Setter_ be capitalized. In fact, make them bold or italic or even inline code snippet to highlight them as important terminology.

Java-Basics.md
```diff
@@ -0,0 +1,442 @@
...
+
+When you write a Class without any constructor, then Java assumes it has a default constructor :
```
alayek: Same as before; this time the word is _Class_

Java-Basics.md
```diff
@@ -0,0 +1,442 @@
...
+
+	public Car() {
+		name = "Tesla";
+	}
+}
+```
+
+Then, when calling `new Car()`, the variable `name` will get auto-initialized to `"Tesla"`.
+
+Like what Constructors can do? [Dive into it more here](#TODO).
```
alayek: Constructors -> `Constructor`s

Java-Basics.md
```diff
@@ -0,0 +1,442 @@
...
+
+### Methods
+
+`getName()` and `getManufacturerName()` are two "Getter" methods we have used here. Notice unlike JavaScript, we **have** to define the return type of any method we write, otherwise it will fail at compile time. If you do not want a method to return anything, use `void` return type.
```
alayek: > Notice unlike JavaScript,

should be

> Notice, unlike JavaScript,

alayek: Link this warning with the info at the top, that Java is a strongly typed language.

Java-Basics.md
```diff
@@ -0,0 +1,442 @@
...
+	public void changeName() {
+		name = "Tesla";
+	}
+}
+```
+As with any other language, methods (or functions, if you are here from JS world) are used often for their modularity and reusability. [Read more about them here](#TODO).
+
+## Inheritance
+
+So great you have successfully created a Car class. But, wait, isnt Tesla cars are supposed to be Electric variants? I want a Electric car class, but it also should have the properties of the original `Car` class.
```
alayek: > Isnt Tesla cars are supposed[...]

should be

> Aren't Tesla cars supposed to be

Also, no need to capitalize _electric_

alayek: > I want a Electric car class, but it also should have the properties of the original `Car` class.

should be

> I want an electric car class, but it also should have the properties of the original `Car` class.

Java-Basics.md
```diff
@@ -0,0 +1,442 @@
...
+
+Solution : Inheritance. Java provides a neat way to "inherit" parent properties :
```
alayek: Inheritance is a domain specific term. Make it either bold or inline code snippet.

Java-Basics.md
```diff
@@ -0,0 +1,442 @@
...
+	}
+}
+
+public class ElectricCar extends Car {
+	public ElectricCar(String name, String man) {
+		super(name, man);
+  }
+
+	public void charge() {
+	 System.out.println("Charging ...");
```
alayek: Indent it properly with equal number of spaces, as there are in the earlier code snippets in the above lines.

Java-Basics.md
```diff
@@ -0,0 +1,442 @@
...
+ElectricCar modelS = new ElectricCar("Model S","Tesla");
+// prints Tesla
+System.out.println(modelS.getName());
+// prints Charging ...
+modelS.charge();
+```
+
+See here that the class `ElecticCar` inherits or `extends` the public methods from `Car` class, as well as has its own methods and properties. Cool way to pass on information!
+
+Also notice the usage of [super]() keyword here. Since our `Car` class had a constructor, so we have to initialize that constructor from the child class as well. We do that using the `super` keyword. Read more about [Inheritance here](#TODO).
```
alayek: Where is the link to super? The link field is empty.

Java-Basics.md
```diff
@@ -0,0 +1,442 @@
...
+		return name;
+	}
+  	// Getter method
+	public String getManufacturerName() {
+		return manufacturerName;
+	}
+}
+```
+
+So interface basically bounds you to a contract to follow, where you must _implement_ all the methods. If you don't, the compiler will complain! Know more about the [awesome power of Inheritance here](#TODO).
```
alayek: > bounds you

should be

> binds you

Java-Basics.md
```diff
@@ -0,0 +1,442 @@
...
+
+For e.g: 
+```java
+int x = 10;
+int y = (x == 10) ? 5 : 9; <-- y will equal 5 since the expression x == 10 evaluates to true
+```
+
+
+###### The `instanceof` operator
+This is a special operator that allows the developer to check whether or not a particular object is in the hierarchy of particular Class/Interface.
```
alayek: Not clear what you are saying here, at least to a beginner.

An object can be part of the hierarchy, irrespective of where in hierarchy it is. It could be above it or below it.

But, only `vehicle instanceof Car` would return `true`. Not `Car instanceof vehicle`.

alayek: Also mention _polymorphism_ here.

Java-Basics.md
```diff
@@ -0,0 +1,442 @@
...
+ 
+ //assuming vehicle is an instance of Class `Car` the expression inside the 'if' will  return true
+ if( vehicle instanceof Car ) {
+   //do something if vehicle is a Car
+ }
+```
+
+## Control Flow
+
+Control flow statements help a developer to take `decisions` based on values. Primarily, Java has the following constructs for control flow:
```
alayek: No, it helps an interpreter or the machine itself, to make a decision at runtime, in a dynamic manner.

Java-Basics.md
```diff
@@ -0,0 +1,442 @@
...
+           break;
+           default :
+           soundNormal();
+       }
+```
+    Note: The `default` case runs when none of the `case` matches. Remember that when a case has no `break` statement, it `falls through` to the next case and will continue to the subsequent `cases` till a `break` is encountered.
+    
+* `nested statements`
+
+Any of the previous control flows can be nested. Which means you can have nested `if`,`if..else` and `switch..case` statements. i.e., you can have any combination of these statements within the other and there is no limitation to the depth of `nesting`.
```
alayek: Provide an example.

Java-Basics.md
```diff
@@ -0,0 +1,442 @@
...
+    
+    
+## Loop constructs
+
+
+Remember the guy (or gal) who goes on and on about something? Well, that's a `loop` right there for you. In `Java`, a developer can represent the action of  `doing something repeatedly` using the following constructs:
```
alayek: No, this is derogatory.

Use something like

> Remember that one friend who goes on[...]

Java-Basics.md
```diff
@@ -0,0 +1,442 @@
...
+
+###### `while` Loop
+
+
+The `while` loop executes a group of statements / single statement till a condition evaluates to `true`. Like so:
```
alayek: > Like so:

should be

> For instance:

Java-Basics.md
```diff
@@ -0,0 +1,442 @@
...
+
+```java
+ while( some condition is true ){
```
alayek: `while(some_condition_is_true)`

Basically, don't put those gaps.

Java-Basics.md
```diff
@@ -0,0 +1,442 @@
...
+  //do something
+  //update the condition
+ }
+```
+`Note`: For the `while` loop to start executing, you'd require the condition to be true. To exit the loop, you'd require to modify the condition so that it becomes false at some point of time. Otherwise, the loop will continue to execute forever.
```
alayek: Once the loop starts executing, there is no modification!

Rewrite the sentence in a way, that it is clear.

Java-Basics.md
```diff
@@ -0,0 +1,442 @@
...
+```
+
+2. Enhanced `for` loop
+
+Well, this came into existence in Java 5. It helps when you are required to iterate over a list of items and perform some action like so:
+
+```java
+ //assuming nameList is a List of names that are actually Strings
+ for( String name : nameList ){
+  SYstem.out.println( name );
```
alayek: `System.out.println(name)`

Java-Basics.md
```diff
@@ -0,0 +1,442 @@
...
+
+int i=10;
+do{
+	System.out.println("The value of i is " + i);
+	i--;
+}while( i >= 10 );
+
+```
+
+`Remember` : The condition of a `do-while` loop is checked AFTER the code body is executed once.
```
alayek: Make AFTER to **After**

Java-Basics.md
```diff
@@ -0,0 +1,442 @@
...
+```
+
+You can create a String in the following ways:
+
+1. `String str = "I am a String";` //This is a String literal
+2. `String str = new String("I am a String")`; //This is a String Object
+
+You might be thinking: What's the difference between the two?
+
+Well, using the `new` keyword gurantees that a new `String` object will be created and a new memory location will be allocated in the `Heap` memory. You see, Java takes care of memory allocation and collecting unused memory in the background - among other things. However, in this case, it's good to be aware about the difference so that you can write code that can help the JVM make appropriate optimizations.
```
alayek: Link to some basic description of `Heap` memory.

Java-Basics.md
```diff
@@ -0,0 +1,442 @@
...
+ String str3 = new String("This is a string");
+ 
+ System.out.println( str == str2 ); //This prints true
+ System.out.println( str == str3 ); //This prints false
+
+```
+
+The answer is: 2 String objects are created. 
+
+You see, the JVM (and its creators) are pretty smart. They figured that Strings differ mostly in terms of its `content`. When you create a String literal, the JVM internally checks, what is known as `the String pool`, to see if it can find a similar (content wise) String object. If it finds it, it returns the same reference. Otherwise, it just goes ahead and creates a new String object in the pool so that the same check can be performed in the future.
```
alayek: > are pretty smart

 would be

> is pretty smart

alayek:
Great job @trishulpani ! Just squash your commits, and provide the example of nested control blocks as mentioned above. Then @koustuvsinha can mark it as complete.

Rafase282:
@trishulpani if you need help squashing let us know.

raisedadead:
Good to go as per @koustuvsinha [here](https://gitter.im/FreeCodeCamp/Wiki?at=56e720a3618c335373ebc98b)

PR #529, opened 2016-03-06
Update Git Resources list

raisedadead:
Updating Git Resources with list of Git Articles.

### DIFF
--- a/Git-Resources.md
+++ b/Git-Resources.md
@@ -1,13 +1,36 @@
-> Resources for working with Git VCS
+# Resources for working with Git VCS
 
-# Books and Tutorials
+A collection of resources on the [Git VCS](https://git-scm.com).
+
+```
+$ git --distributed-is-the-new-centralized
+```
+*Git is a free and open source distributed version control system designed to handle everything from small to very large projects with speed and efficiency.*
+
+## FreeCodeCamp's Wiki Resources
+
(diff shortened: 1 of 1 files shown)


Rafase282:
Fancy table.

raisedadead:
Umm this confuses me! Are you approving it or rejecting it! :fearful:

raisedadead:
@Rafase282 On a side note however I guess we could come up with a way to hot wire the CamperBot to point to the list? Not sure what should be the format. Do you have any pointers on that? or maybe @ltegman is the expert on everything bot?

Rafase282:
@raisedadead the table is nice but a simple ordered list would achive the same result and make it easier to maintain. However if people want to use the table I wont oppose to it. I would personally prefer:

1. A nice
2. Ordered
3. List
4. Like this.

Rafase282:
@raisedadead currently this is what is shown when you do a `wiki git` [url](https://github.com/freecodecamp/freecodecamp/wiki/challenge-save-your-code-revisions-forever-with-git)

raisedadead:
@Rafase282 I recommend we move to a simpler un-ordered list that's sorted alphabetically, it would be much simpler to add entries that way without making too many changes to the entire list.

And thanks for the guidance on the bot, I'll check and update once I figure out.

PR #532, opened 2016-03-07
Update wiki article about CSS Selectors

pedronfigueiredo:


Rafase282:
See #542

VeerpalB:
Opps, I didn't notice that you had also made a pull request for this! I created #543. Why don't you incorporate some of the stuff from my update into your file, and then I'll close my PR. 

In particular, consider adding:

 A,B              - target all A elements and all B elements
 A~B              - target all B elements that have an A element preceeding it
 A:not(.name)      - Target all A elements that do NOT have the class name
 A::before         - Insert something before the content of the A element
 A::after          - Insert something after the content of the A element
 A::first-child    - Target all A elements that are the first child of their parent

as well as some of the links I mentioned.

pedronfigueiredo:
Ok, will do!

PR #536, opened 2016-03-07
Add Campsite JSON File

LibertysYarn:


raisedadead:
@LibertysYarn Thanks for the PR, I am editing the PR title, we should not include any issue no.s in the commit message or the PR titles.

Adding @hallaathrad for QA.

Rafase282:
It's been over a week. What's the status on this? is it useful? should it be merged? closed?

abhisekp:
Would be nice to add latitude and longitude info instead of null (*optional*)

samosale:
Have a list with all cities(950) - 750 have coordinates, but some are incorect(due to same names of different cities ex. Moscow Russia vs Moscow Usa ) 
https://github.com/samosale/fccCities/blob/master/fccCitiesFull.json

One should fetch data from here https://developers.google.com/maps/documentation/geocoding/intro?csw=1#BYB

Example https://maps.googleapis.com/maps/api/geocode/json?address=Jyv%C3%A4skyl%C3%A4,+Finland

hallaathrad:
So how would y'all say we proceed about this?

samosale:
I will fetch coords from google api just to grab some time.

alayek:
It's been 16 days of inactivity on this PR. What to do here?

@Rafase282 @raisedadead please take a look.

Rafase282:
Feel free to re-open when there is something new.

PR #537, opened 2016-03-07
Fix links in CONTRIBUTING guidelines

raisedadead:
Updated the links, were leading to 404s.


raisedadead:
@Rafase282 they are broken when I visit while creating a new issue:
image

which goes to https://github.com/FreeCodeCamp/wiki/blob/master/.github/CONTRIBUTING.md

where all the links don't work.

raisedadead:
Its the way I guess github links the Contributing guide

Rafase282:
Merging for now. We will have to check whether to keep using relative links or go back to hard links.

PR #550, opened 2016-03-08
Update Python Resources with fragmented URLs

alayek:
I couldn't test this locally. I guess we can test it only if it's live.

@Rafase282 @koustuvsinha FYI.


Rafase282:
@alayek they worked from the preview.

PR #575, opened 2016-03-14
Adding Ruby introduction

Sayanc93:


Rafase282:
You can do ```ruby for syntax highlight.

Rafase282:
Other than that it looks good

Sayanc93:
@Rafase282 - Hey thanks for the suggestion, but I was trying to be consistent with Python introduction...it followed the same pattern, should i change it to ``` or let it be?

raisedadead:
@Sayanc93 What @Rafase282 Meant is since the code snippets are in ruby you need to add them as:


\`\`\`ruby
your code here...
\`\`\`

PR #498, opened 2016-03-03
Known-Issue-With-Codepen.md

micahbales:
I spent about a week trying to debug a problem that ended up being an issue with a Chrome extension conflicting with Codepen. This file is meant to provide a list of known issues with Codepen, to reduce frustration in the future.

PR #585, opened 2016-03-16
Update issues

Rafase282:
See https://github.com/FreeCodeCamp/wiki/pull/498 for history.


abhisekp:
# :+1: LGTM

Would have been nice to add any alternatives to CodePen. (*optional*)

Rafase282:
I'm not sure what would be a good alternative and i was not sure about adding the link to that imgur replacement as I don't want to endorse anything under FCC's name.

PR #587, opened 2016-03-16
Updates format

Rafase282:
#585


abhisekp:
# :+1: LGTM

> Would be nice to add any alternatives to CodePen. (*optional*)

PR #619, opened 2016-03-20
Create a wiki article on installing CLI tools

mAzurkovic:
This article will explain to readers how to install widely used CLI tools for web development.

### DIFF
--- a/Installing-Necessary-CLI-Tools.md
+++ b/Installing-Necessary-CLI-Tools.md
@@ -1,0 +1,35 @@
+# Installing CLI Tools Developers Should Not Live Without (Linux and Mac)
+
+This article is a short description on how to install key command line utilities that developers use everyday on Macintosh and Linux environments. The main CLI tools that will be shown are: Homebrew (Mac), Node and NPM, Bower, Gulp and Grunt.
+
+### Installing Homebrew (Macintosh Systems)
+
+Homebrew is ‘The missing package manager for OS X’. It is a great tool for downloading and installing packages straight from your command line. This is not needed on Linux distributions because they already have package managers installed by default such as `apt-get` or `pacman`. To install Homebrew, simply paste the following command in your terminal:
+* `/usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"`
+
+### Installing NPM
+
+`NPM`, or Node Package Manager, is another useful package manager for downloading mostly web tools. Downloading `NPM` will also install the Node.js framework.
+
+##### Mac:
+* Using `Homebrew` type: `brew install node` and `NPM` should have been installed.
+
+##### Linux:
+* Using `apt-get` first type: `curl -sL https://deb.nodesource.com/setup_4.x | sudo -E bash -`, then `sudo apt-get install nodejs`
+* Using `pacman` type: `pacman -S nodejs npm`
+* Using `yum` type: `sudo yum install nodejs npm`
+
+### Installing Bower
+Bower is a popular package manager that allows you to install front-end libraries. You can install it on Linux and Macintosh systems using `npm` with the following command:
+* `npm install bower`
+(Same command for both Linux and OS X)
+
+### Installing Gulp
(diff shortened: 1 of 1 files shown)


Installing-Necessary-CLI-Tools.md
```diff
@@ -0,0 +1,35 @@
...
+* `/usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"`
+
+### Installing NPM
+
+`NPM`, or Node Package Manager, is another useful package manager for downloading mostly web tools. Downloading `NPM` will also install the Node.js framework.
+
+##### Mac:
+* Using `Homebrew` type: `brew install node` and `NPM` should have been installed.
+
+##### Linux:
```
alayek: Node JS should not be installed via package manager. A lot of programs that depend on Node JS, requires Node JS to be invoked as `$ node` in terminal. So, add some guide to update symlinks properly.

The recommended way to install Node JS is either compile directly from source; or install `nvm` first, then go with `nvm use <version_number>`. Given how quickly Node JS has been releasing new versions, it's better to go with second one to stay on top of things, even when you need LTS versions.

Installing-Necessary-CLI-Tools.md
```diff
@@ -0,0 +1,35 @@
...
+* Using `apt-get` first type: `curl -sL https://deb.nodesource.com/setup_4.x | sudo -E bash -`, then `sudo apt-get install nodejs`
+* Using `pacman` type: `pacman -S nodejs npm`
+* Using `yum` type: `sudo yum install nodejs npm`
+
+### Installing Bower
+Bower is a popular package manager that allows you to install front-end libraries. You can install it on Linux and Macintosh systems using `npm` with the following command:
+* `npm install bower`
+(Same command for both Linux and OS X)
+
+### Installing Gulp
```
alayek: It would be a good idea to mention it to the reader that Gulp and Grunt both are task runners, but one focuses on config while other focuses on code. 

So if they have one, they might not need the other.

Installing-Necessary-CLI-Tools.md
```diff
@@ -0,0 +1,35 @@
...
+`Gulp` is a framework and CLI tool that makes development faster and more enjoyable by automating tasks developers find themselves doing over, and over again. Also, you can install `Gulp` through `npm`:
+* `npm install --save-dev gulp`
```
alayek: This is assuming you already have a `package.json` in the same directory, and the user already has a project set up.

Installing-Necessary-CLI-Tools.md
```diff
@@ -0,0 +1,35 @@
...
+
+### Installing Grunt
+`Grunt` is another popular automation tool similar to `Gulp`. To install it, use `npm` again:
+* `npm install -g grunt-cli`
```
alayek: This command most likely won't work, because you didn't use `sudo`. There are ways to make `npm` global installations work without `sudo`. Kindly at least link to some external guide that helps you set up do that; or add those instructions as well.

Installing-Necessary-CLI-Tools.md
```diff
@@ -0,0 +1,35 @@
...
+
+And that's that! With these tools your development process and content will be enjoyable and efficient. As you can see, the main tool to install is `npm` and it allows you to install many other great web oriented CLI tools. Happy Hacking! :+1:
```
alayek: FreecodeCamp recently expanded its back-end curriculum to have support for languages other than JS - such as Ruby, Python, Java, Go etc.

In light of this, it would be a great idea to mention that these CLI tools are for front-end devs, or back-end JS devs. It needs to be clear that these **DO NOT** contain instructions to install Python or Ruby Web-Dev CLI tools. 

You might want to change the title of the guide.

Installing-Necessary-CLI-Tools.md
abhisekp: If nodejs/npm is installed in `sudo` mode, then the global packages must also be installed in `sudo` mode.

alayek:
@mAzurkovic we don't know if you are still working on this. Would you be able to implement the suggestions and update the PR soon?

alayek:
@FreeCodeCamp/wiki-mods do not remove the labels - we need to raise a PR to fix some stuff in the articles.

Issue #534, opened 2016-03-07
Wiki Request - New Articles on Java are required


koustuvsinha:
This is an epic issue to track the development of Java Wiki articles. The target is to have smaller articles on individual topics so that they can be easily referenced by @camperbot . 

These are our requirements. Feel free to add comment on articles which are needed and I will update it here. If you want to work on a specific article, comment here and let others know. Also, refer this issue in subsequent PR's for easy tracking.

- [x] [Home Page](https://github.com/FreeCodeCamp/FreeCodeCamp/wiki/Java)
- [x] [Introduction](https://github.com/FreeCodeCamp/FreeCodeCamp/wiki/Java-Introduction)
- [ ] Basics - In progress #526 
- [ ] Data Types
- [ ] Control Statements
- [ ] Loops
- [ ] Constructors
- [ ] Inheritance
- [ ] Interfaces
- [ ] Abstract Classes
- [ ] Packages
- [ ] Lambda Expressions
- [ ] Project Structure
- [ ] Nested Inner Anonymous Class
- [ ] Unit Testing in Java
- [ ] Web Development with Java
- [ ] Getting started with Spring
- [ ] Spring Installation
- [ ] Planning a Spring Web App
- [ ] Spring Model Layer
- [ ] Spring Routes
- [ ] Spring Views
- [ ] Spring Templates
- [ ] Testing in Spring
- [ ] Spring Refactoring & Improvements
- [ ] Java More Resources

@Rafase282 @raisedadead please add relevant labels as per other issues. @alayek have a look.

alayek:
Great job @koustuvsinha !

ScrumUser1:
I would like to contribute to the loop article for java. Can we talk more about this so i know how you would want it to be?

alayek:
@ScrumUser1 yes! We communicate via gitter. Kindly contact me or @koustuvsinha via hitter, and one of us would discuss with you how the article would be.

koustuvsinha:
@ScrumUser1 can you work on a different article as another user has already submitted a PR on loops in #560 ? If you need any help please feel free to contact me in [Gitter](https://gitter.im/koustuvsinha).

trishulpani:
I'd like to take up Constructors. Let me know if anyone else is looking into it.

Shaif00:
Hii, I'm writing data types and control Statements, will create a pull req soon.

alayek:
@koustuvsinha FYI. Let me know if I need to assign something.

[...]

PR #623, opened 2016-03-21
Java data types

Shaif00:
Java Data types. See if this is all right. #534

### DIFF
--- a/Java-Data-Types.md
+++ b/Java-Data-Types.md
@@ -1,0 +1,183 @@
+# Java Data-Types
+
+Java is a strongly typed language. This means that, in Java, each data type has its own strict definition. There are no implicit data type conversions when any conflicts occur between the data types. Any change in data types should be explicitly declared by the programmer.
+ 
+Java defines 8 simple data types : `byte`, `short`, `int`, `long`, `char`, `float`, `double` and `boolean`. 
+
+They are divided into the following categories:
+ 
+#####Integers      
+#####Floating Point Numbers      
+#####Characters      
+#####Boolean Type
+<br />
+The details of each of the data types is given below : 
+
+###Integers:
+
+These are of four types: `byte`, `short`, `int`, `long`. It is important to note that these are signed positive and negative values.  Signed integers are stored in a computer using [2's complement](http://www.ele.uri.edu/courses/ele447/proj_pages/divid/twos.html). It consist both negative and positive values but in different formats like `(-1 to -128)` or `(0 to +127)`. An unsigned integer can hold a larger positive value, and no negative value like `(0 to 255)`. Unlike C++ there is no unsigned integer in Java.
+
+####byte:
+
+Byte data type is an 8-bit signed two's complement integer.
+
+```
+Wrapper Class: Byte
+
+Minimum value: -128 (-2^7)
+
+Maximum value: 127 (2^7 -1)
+
+Default value: 0
+
+Example: byte a = 10 , byte b = -50
+```
+
+
+#### short:
+
+Short data type is a 16-bit signed two's complement integer.
+
+```
+Wrapper Class: Short
+
+Minimum value: -32,768 (-2^15)
+
+Maximum value: 32,767 (2^15 -1)
+
+Default value: 0.
+
+Example: short s = 10, short r = -1000
+```
+
+#### int:
+
+int data type is a 32-bit signed two's complement integer. It is generally used as the default data type 
+for integral values unless there is a concern about memory.
+
+```    
+Wrapper Class: Integer
+
+Minimum value: (-2^31)
+
+Maximum value: (2^31 -1)
+
+The default value: 0.
+
+Example: int a = 50000, int b = -20
+```
+
+#### long:
+
+Long data type is a 64-bit signed two's complement integer.
+
+```    
+Wrapper Class: Long
+
+Minimum value: (-2^63)
+
+Maximum value: (2^63 -1)
+
+Default value: 0L.
+
+Example: long a = 100000L, long b = -600000L; 
+
+By default all integer type variable is "int". So long num=600851475143  will give an error.
+But it can be specified as long by appending the suffix L (or l) 
+```   
+    
+### Floating­ Point​:
+
+These are also called real numbers and are used for expressions involving fractional precision. These are of two types: `float`, `double`. Float is actually avoided in case of precise data such as currency or research data.
+
+#### float:
+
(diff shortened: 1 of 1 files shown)


Java-Data-Types.md
```diff
@@ -0,0 +1,148 @@
+# Java data-Types
+
+Java is a strongly ­typed language. This means that, in Java, each data type has its own strict definition. There are no implicit data type conversions when any conflicts occur between the data types. Any change in data types should be explicitly declared by the programmer.
+ 
+Java defines 8 simple data types. They are divided into the following categories:
```
alayek: When you are saying 8 data types, using a cardinal number; better to just list down the 8 types, instead of creating sub-classes.

Java-Data-Types.md
```diff
@@ -0,0 +1,148 @@
...
+ 
+Characters
+ 
+Boolean Type
+ 
+The details of each of the data types is given below
+ 
+### Integers:
+
+These are of four types: byte, short, int, long. It is important to note that these are signed positive and negative values. Java does not support positive­ only numbers.
```
alayek: You are writing in markdown, so use the power of that judiciously.

> These are of four types: byte, short, int, long

should be

> These are of four types: `byte`, `short`, `int`, `long`

Also note that you are listing only the  primitive types. I think you should mention the proper wrapper classes along with them as well.

Java-Data-Types.md
```diff
@@ -0,0 +1,148 @@
...
+
+#### byte:
+
+    Byte data type is an 8-bit signed two's complement integer.
+
+    Minimum value: -128 (-2^7)
+
+    Maximum value: 127 (inclusive)(2^7 -1)
```
alayek: No need to write inclusive. If it cannot reach that value, it won't be its maximum.

Java-Data-Types.md
```diff
@@ -0,0 +1,148 @@
...
+
+    Default value: 0
+
+    Example: byte a = 10 , byte b = -500
```
alayek: I think you just said that minimum value cannot be less than -128. What happens if you assign -500 to a `byte` variable?

Java-Data-Types.md
```diff
@@ -0,0 +1,148 @@
...
+
+#### short:
+
+    Short data type is a 16-bit signed two's complement integer.
```
alayek: See, this is what sets a good writer apart from a moderate writer. It's the amount of information you throw at people within a given amount of time or breadth of an article.

These articles should not put off beginners. You are using the term "two's complement integer" without any explanation as to what they mean.

At the very least, link to a Wikipedia article or some lucid article as to what they are. Not everyone would have an understanding of that.

Java-Data-Types.md
```diff
@@ -0,0 +1,148 @@
...
+
+    Minimum value: -32,768 (-2^15)
+
+    Maximum value: 32,767 (inclusive) (2^15 -1)
```
alayek: Same as above. No need to write inclusive.

Java-Data-Types.md
```diff
@@ -0,0 +1,148 @@
...
+
+    Default value: 0.
+
+    Example: short s = 10, short r = -1000
+
+#### int:
+
+    Int data type is a 32-bit signed two's complement integer. Int is generally used as the default data type
```
alayek: > Int

There is no such thing.

Use `int`

Java-Data-Types.md
```diff
@@ -0,0 +1,148 @@
...
+
+    Long data type is a 64-bit signed two's complement integer.
+
+    Minimum value: (-2^63)
+
+    Maximum value: (2^63 -1)
+
+    Default value: 0L.
+
+    Example: long a = 100000L, long b = -600000L
```
alayek: You are using `L` at the end of a number to denote a  `long`. You and I would get it. But not everyone reading this article would get it.

Clearly explain that `L` is not a special number, rather a compiler directive. And what it directs the compiler to do.

Java-Data-Types.md
```diff
@@ -0,0 +1,148 @@
...
+
+
+### Floating­ Point​:
+
+These are also called real numbers and are used for expressions involving fractional precision. These are of two types: float, double. Float is actually avoided in case of precise data such as currency or research data.
+
+#### float:
+
+    Float data type is a single-precision 32-bit IEEE 754 floating point.
```
alayek: Link to some IEEE paper or Wikipedia article on what IEEE 754 spec is.

Java-Data-Types.md
```diff
@@ -0,0 +1,148 @@
...
+
+    Float is mainly used to save memory in large arrays of floating point numbers.
+
+    Default value: 0.0f.
```
alayek: Same as above. You are using `f` at the end of the number literal, to direct the compiler to store it as floating point. Mention that clearly.

Also note that if you try to do `float myFloat = 7.8`, it would give a compiler error. A lot of beginners get put off by this.

Java-Data-Types.md
```diff
@@ -0,0 +1,148 @@
...
+
+    Example: float f1 = 24.5
```
alayek: As said above, this would give an error.

Java-Data-Types.md
```diff
@@ -0,0 +1,148 @@
...
+    This data type is generally used as the default data type for decimal values, generally the default choice.
+
+    Default value: 0.0d.
+
+    Example: double d1 = 123.400778
+
+ 
+### Character:
+
+We use this data type to store characters. This is not the same as the char in C/C++. Java uses an UNICODE, internationally accepted character set. Char in Java is 16­bits long while that in C/C++ is 8­bits.
```
alayek: > Java uses an UNICODE

should be

> Java uses UNICODE

Also, use "a", not "an".

alayek:
@koustuvsinha @abhisekp please take a look. I am in a hurry, so I might have missed a few things.

@Shaif00 thanks for the article. Kindly implement suggestions to make the article better.

Shaif00:
@alayek made changes.

Rafase282:
@shaduk I sould suggest using an alternative to the block quotes using plain backticks and perhaps some more bold keywords in the documents but other than that it looks good.

Rafase282:
@Shaif00 I pointed the wrong person but just in case, I was talking to you with regards to the markdown.

Shaif00:
@Rafase282  commits squashed

alayek:
@Shaif00 great job! I hope you keep this stamina going.

@Rafase282 I am okay with this. But if he cannot implement all suggestions right now, I will raise a PR later to improve on it.

@koustuvsinha if you get time, just take a look.

PR #632, opened 2016-03-23
Update to follow new template and add one more solution

cloudb:
I have changed the entry to correspond with those made with the new template and I have also added a new intermediate solution with explanation.

### DIFF
--- a/Algorithm-Return-Largest-Numbers-in-Arrays.md
+++ b/Algorithm-Return-Largest-Numbers-in-Arrays.md
@@ -104,10 +104,10 @@ Here we're passing `null` as the *context* of the `Function.apply` method as `Ma
 *Bien noté*? :neutral_face:
 
 **Reference:-**
-- http://devdocs.io/#q=js+Math+max
-- http://devdocs.io/#q=js+Array+map
-- http://devdocs.io/#q=js+Function+apply
-- http://devdocs.io/#q=js+Function+bind
+- [Math.max on DevDocs](http://devdocs.io/#q=js+Math+max)
+- [Array.map on DevDocs](http://devdocs.io/#q=js+Array+map)
+- [Function.apply on DevDocs](http://devdocs.io/#q=js+Function+apply)
+- [Function.bind on DevDocs](http://devdocs.io/#q=js+Function+bind)
 
 ### :trophy: Credits:
 If you found this page useful, you can give thanks by copying and pasting this on the main chat: 
--- a/Algorithm-Return-Largest-Numbers-in-Arrays.md
+++ b/Algorithm-Return-Largest-Numbers-in-Arrays.md
@@ -1,13 +1,15 @@
-# Explanation:
+:triangular_flag_on_post: Remember to use [**`Read-Search-Ask`**](http://github.com/FreeCodeCamp/FreeCodeCamp/wiki/How-to-get-help-when-you-get-stuck) if you get stuck. Try to pair program. :busts_in_silhouette: Write your own code. :memo:
+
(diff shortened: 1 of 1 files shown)


Rafase282:
@cloudb we are not taking new solutions until we implement the new template, but given the changes you made I think we can accept this on. I would change your reference links to have just titles like the ones on the intermediate one.

cloudb:
@Rafase282 I did not realise that you guys were not taking solution changes.  I can wait until the new template is ready before continuing with this amendment. Is the layout I used adhering to the new template or are there further changes on the horizon.  I'm just looking to contribute as I work through the syllabus.

RE: you reference links sugestion

I'm a little confused, the intermediate solution (and links) is the solution I am adding.  The changes I made to the basic and advanced solutions just involved formatting with respect to the new template rules.  Or do you mean I should be changing my link display text from "Map on MDN" to "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map"?

Rafase282:
@cloudb The one you used is the one we want to implement with all of the others and we could use your help on that task.

About the links, sorry for the confusion. It is the other way around, could you make the others display like [Map on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map). instead of showing the full link.

cloudb:
That makes sense.  I'll make the changes. How does this work? Do i need to make a new pull request or can I just commit to my fork and you will accept both sets of changes?

I'm more than happy to help move to the new template as I work through the algorithms,

Algorithm-Return-Largest-Numbers-in-Arrays.md
```diff
@@ -85,14 +103,20 @@ i.e. `Math.max.apply(null, [9, 43, 20, 6]);` would invoke something like a `Max.
...
-- http://devdocs.io/#q=js+Array+map
-- http://devdocs.io/#q=js+Function+apply
-- http://devdocs.io/#q=js+Function+bind
-
-# Credits:
-If you found this page useful, you can give thanks by copying and pasting this on the main chat: **`Thanks @Rafase282 @abhisekp @Hallaathrad for your help with Algorithm: Return Largest Numbers in Arrays`**
-
-> **NOTE:** Please add your username only if you have added any **relevant main contents** to the wiki page. (Please don't remove any existing usernames.)
+- [Math.max on DevDocs](http://devdocs.io/#q=js+Math+max)
+- [Array.map on DevDocs](http://devdocs.io/#q=js+Array+map)
```
Rafase282: [Math.max](Math.max)
[Array.map](js-Array-prototype-map)

Algorithm-Return-Largest-Numbers-in-Arrays.md
```diff
@@ -85,14 +103,20 @@ i.e. `Math.max.apply(null, [9, 43, 20, 6]);` would invoke something like a `Max.
 
 *Bien noté*? :neutral_face:
```
Rafase282: Please remove this.

Rafase282:
@cloudb I would suggest a new pull request per challenge to keep QA more clean and focus. Now all you have to do is squash after the minor adjustments I suggested. Something you should know is to use internal wiki links whenever possible instead of external ones.

You can search [here](https://github.com/FreeCodeCamp/FreeCodeCamp/wiki) if you don't know if we already have an article or not. These would use relative links meaning `[Array.map](js-Array-prototype-map)` instead of `[Array.map](https://github.com/FreeCodeCamp/FreeCodeCamp/wiki/js-Array-prototype-map)`

PR #671, opened 2016-04-04
All of the cities are belong to us :)

hallaathrad:


Rafase282:
Does this means you have finished with that long list?

PR #673, opened 2016-04-04
Update Algorithm-Caesars-Cipher.md

The-Seth:


Rafase282:
We are taking only one solution per category. If you want yours to replace the existing one then provide an explanation as to why yours is better or/and why yours covers the save and teaches even more.

The-Seth:
My algorithm does not require treating the first half and second half of the alphabet differently, unlike all other algorithms listed. Once the character is recognized as belonging to the alphabet, a single statement takes care of the conversion.

alayek:
@The-Seth In fairness, this is the solution I used when I was solving this. But your solution suffers from **magic number** problem. Also, what happens when the letters are lowercase?

The-Seth:
I agree it is not *the best* solution out there for a cipher problem, but then I would also say neither is any of the other one's posted. This is genuinely a question from me (I am not a computer science engineer nor a programmer at job, I had to google for what the magic number problem is) - don't other solutions have the magic number problem too?

I am not trying to nitpick on anyone else's efforts but just like other solutions posted, this solution works for the challenge at hand.

alayek:
I understand you must be feeling frustrated over this PR being closed. I haven't checked the original solution on the wiiki before you submitted this PR.

So, I decided to raise another PR after your PR was closed: https://github.com/FreeCodeCamp/wiki/pull/677

Feel free to review it. I am planning to add your name in the list of contributors there. Let me know if you are okay with that. I don't want to use your name without your permission.

The-Seth:
I was not really frustrated :) just curious.

I checked your PR #677 and it looks good to me, thank you for that. Even I am learning to become a better programmer, so I am definitely in support of your approach of assigning constants dynamically to variables before using them, because then it needs to be done only once.

And I am okay with my name being in the list.

PR #677, opened 2016-04-05
Fix Algorithm Caesars Cipher

alayek:
I noticed some magic numbers in the solution, and decided to fix that. It's not wrong to use numbers like 65 or 90 or 78 in a solution. But, it's a good practice not to use those all over the place - rather create the constant values dynamically at one place and reuse them.

Also, noticed that the solution doesn't use modular arithmetic. Thanks @The-Seth for pointing that out.

### DIFF
--- a/Algorithm-Caesars-Cipher.md
+++ b/Algorithm-Caesars-Cipher.md
@@ -35,19 +35,27 @@ e.g. 'A' ↔ 'N', 'T' ↔ 'G'.
 ```js
 function rot13(str) {
     var nstr="";
+    /**
+    * It's not a good idea to use numbers other than 0 or 1 in your code
+    * So we will dynamically obtain ASCII values of A, Z and mid-point
+    **/
+    var VALUE_OF_A = "A".charCodeAt(0); // ASCII value of A, which is 65
+    var VALUE_OF_Z = "Z".charCodeAt(0); // ASCII value of Z, which is 90
+    var LENGTH_OF_ALPHABET = VALUE_OF_Z - VALUE_OF_A + 1; // this is 26
+    var VALUE_OF_MID_POINT = VALUE_OF_A + LENGTH_OF_ALPHABET/2; // ASCII value of halfway letter in alphabet, which is 78
     for(var i=0; i<str.length; i++){
         //Checks if character lies between A-Z
-        if(str.charCodeAt(i) < 65 || str.charCodeAt(i) > 90) {
-            nstr += String.fromCharCode(str.charCodeAt(i));
+        if(str.charCodeAt(i) < VALUE_OF_A || str.charCodeAt(i) > VALUE_OF_Z) {
+            nstr += str[i];
             continue;
         }
         //N = ASCII 78, if the character code is less than 78, shift forward 13 places
-        if(str.charCodeAt(i) < 78){
-            nstr += String.fromCharCode(str.charCodeAt(i) + 13);
+        if(str.charCodeAt(i) < VALUE_OF_MID_POINT){
+            nstr += String.fromCharCode(str.charCodeAt(i) + LENGTH_OF_ALPHABET/2);
         }
         else{
             //Otherwise shift the character 13 places backward
-            nstr += String.fromCharCode(str.charCodeAt(i) - 13);
+            nstr += String.fromCharCode(str.charCodeAt(i) - LENGTH_OF_ALPHABET/2);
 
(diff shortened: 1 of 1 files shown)


Algorithm-Caesars-Cipher.md
```diff
@@ -35,19 +35,27 @@ e.g. 'A' ↔ 'N', 'T' ↔ 'G'.
...
+    **/
+    var VALUE_OF_A = "A".charCodeAt(0); // ASCII value of A, which is 65
+    var VALUE_OF_Z = "Z".charCodeAt(0); // ASCII value of Z, which is 90
+    var LENGTH_OF_ALPHABET = VALUE_OF_Z - VALUE_OF_A + 1; // this is 26
+    var VALUE_OF_MID_POINT = VALUE_OF_A + LENGTH_OF_ALPHABET/2; // ASCII value of halfway letter in alphabet, which is 78
     for(var i=0; i<str.length; i++){
         //Checks if character lies between A-Z
-        if(str.charCodeAt(i) < 65 || str.charCodeAt(i) > 90) {
+        if(str.charCodeAt(i) < VALUE_OF_A || str.charCodeAt(i) > VALUE_OF_Z) {
             nstr += String.fromCharCode(str.charCodeAt(i));
```
abhisekp: The following 
```js
nstr += String.fromCharCode(str.charCodeAt(i));
```
is same as
```js
nstr += str[i];
```

----
> this correction can be made in a new commit for improving the algo

abhisekp:
@alayek 
After analyzing all the solutions, I conclude the following
- The **existing Advanced solution** should replace the **existing Basic solution**.
- The new **Alternative Basic solution** is an Advanced solution and should be added in **Advanced category**.

Make different commits (*in this order*) for 
1. Moving/Replacing existing solutions
2. Optimizing any existing solution
3. Updating the CONSTANT values in the solution (*optional as I don't think this is necessary*) 
4. Adding new solution

> I don't think replacing the CONSTANT values is important because the algorithms are not for production ready code and are very specifically for our FCC algorithm solution section. So refactoring all the CONSTANTs is unnecessary.

abhisekp:
But also refactoring CONSTANTs is necessary for larger code solutions for better understanding, so I guess, :+1:

alayek:
Ok, I will update the PR with those commits. I agree that being from a non-CS background, noting the circular buffer pattern would be a bit daunting - hence we should move that to advanced.

And you are right about the constants - we are not doing this for production ready code. I guess I would leave those as is. And add a line somewhere about using constants for better maintainability and readability.

I am waiting for @SaintPeter to chime in, because he was one of the original creator of this; and he should have a say in this.

SaintPeter:
I will take @abhisekp's word with regards to the rearranging.

I do like making the "magic numbers" into generated numbers.  That's pretty slick and improves readability.

My only "complaint" is that no one did a functional solution - it seems like there should always be a really nice tight functional solution.  

Amusingly:  Originally this was a checkpoint and my solution used a giant lookup table.  I never actually solved it "properly".  :smile:

alayek:
@SaintPeter ok, let me jot up a functional solution and have it reviewed in [Wiki chatroom](https://gitter.im/FreeCodeCamp/Wiki). Then I would add in the PR.

abhisekp:
@alayek I've a functional solution if you want to add it to the advanced solution list
```js
  return str.replace(/[A-Z]/g, (L) => String.fromCharCode(65 + L.charCodeAt(0) % 26));
}
```

Here is my complete list https://gist.github.com/abhisekp/a891a7318160c989f251

> NOTE: Change function names accordingly as per the new function names

Algorithm-Caesars-Cipher.md
```diff
@@ -68,6 +76,43 @@ ascii code.
 range when subtracted by 13 so we'll add 26 (number of letters in English alphabets) to it so that after A it'll go back to Z.
 e.g. M(77) ↔ 77-13 = 64(Not an English alphabet) +26 = 90 ↔ Z(90)
 
+It is quite possible to replace the `if-else` logic with a simple modulo operation, because of the cyclic nature of English alphabet.
+
+So, our final code could very well look like as follows.
```
theflametrooper: ~~So, our final code could very well look like as follows.~~ So, our final code could very well look like what's show below.

alayek:
I am closing this because this is no longer required. I have created a new PR at #689

PR #689, opened 2016-04-07
Fix algorithm caesars cipher

alayek:
This is a fix to the issues raised in https://github.com/FreeCodeCamp/wiki/pull/677

I have moved and updated the solutions as @abhisekp has mention [here](https://github.com/FreeCodeCamp/wiki/pull/677#issuecomment-205645189).

I have also added functional solution by @abhisekp in the advanced section.

I have not included any constants to replace the magic numbers, because that would be overkill!

### DIFF
--- a/Algorithm-Caesars-Cipher.md
+++ b/Algorithm-Caesars-Cipher.md
@@ -85,7 +85,7 @@ function rot13(str) {
       // A more general approach
       // possible because of modular arithmetic 
       // and cyclic nature of rot13 transform
-      rotCharArray.push((str[x].charCodeAt - 65 + 13) % 26 + 65);
+      rotCharArray.push((str[x].charCodeAt() - 65 + 13) % 26 + 65);
     } else {
       rotCharArray.push(str[x].charCodeAt());
     }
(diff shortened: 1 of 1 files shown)


alayek:
@SaintPeter @Rafase282 please take a look.

Rafase282:
Second solution does not work. See the wiki chat.

Imgur

For the advanced one:

I think this is a bug on FCC but still pointing that out.
Imgur

abhisekp:
@Rafase282 that warning is because FCC editor uses JSHint as the linter and probably, has not set `esversion: 6` in its options somewhere (*i'm not sure about the options thing though*).

I don't think it's required to add `esversion: 6` as it's unrelated to the code and is an optional thing which is related to the editor the code is executed in.

Rafase282:
@abhisekp yeah i was told about that which is why I pointed to the chat, I didn't remove it so in the future we have a reference if we use the code and campers get concerned about.

The bug part is because the intermediate code is not working.

SaintPeter:
@abhisekp Would you be willing to write up a more detailed explanation for your super-tight functional solution?

Rafase282:
@alayek what's the status of this PR?

abhisekp:
@Rafase282 I need some time to think about the logic I used and try to explain it in most succinct way.

abhisekp:
<h1 align="center"><a href="https://github.com/FreeCodeCamp/wiki/pull/689#issuecomment-207077267">TODO</a></h1>

Write an easy to understand explanation of :rotating_light: **Advanced Solution**.

PR #690, opened 2016-04-07
Add git behind proxy article

alayek:
Added article on how to use git behind proxy server.

A lot of our campers are accessing GitHub in their colleges, or workplace. This would surely serve as a reference for them.

PR #691, opened 2016-04-07
Add NPM Behind Proxy Article

alayek:
Similar to https://github.com/FreeCodeCamp/wiki/pull/690, except for `npm`.

### DIFF
--- a/NPM-Behind-Proxy.md
+++ b/NPM-Behind-Proxy.md
@@ -1,0 +1,40 @@
+# NPM Behind A Proxy Server
+
+### Use-cases
+
+You might need to modify `npm install` commands that access remote repositories ([npmjs](https://www.npmjs.com/), for example) for installing Node JS modules; if your internet access is through a [proxy server](https://en.wikipedia.org/wiki/Proxy_server).
+
+Proxy servers are common in college and business type environments.
+
+You can [locate your proxy settings](http://www.wikihow.com/Change-Proxy-Settings) from your browser's settings panel.
+
+### Using Proxy with NPM
+
+Once you have obtained the proxy settings (server URL, port, username and password); you need to configure your `npm` configurations as follows.
+
+```bash
+$ npm config set proxy http://<username>:<password>@<proxy-server-url>:<port>
+$ npm config set https-proxy http://<username>:<password>@<proxy-server-url>:<port>
+```
+You would have to replace `<username>`, `<password>`, `<proxy-server-url>`, `<port>` with the values specific to your proxy server credentials.
+
+These fields are optional. For instance, your proxy server might not even require `<username>` and `<password>`, or that it might be running on port 80 (in which case `<port>` is not required).
+
+Once you have set these, your `npm install`, `npm i -g` etc. would work properly.
+
+### When Not to Use
+
+You should not have to use `npm` commands with proxy settings, if either of the following happens:
+
+>- Your system administrator or corporate policy does not allow you to access remote `npm` repositories from NPM-JS, for instance.
+- The remote repository of Node modules in question is not in your machine, but it's within the internal network.
+
+### Unset Proxy Settings
+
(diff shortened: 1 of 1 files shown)


Rafase282:
@SeanFCC @theflametrooper  please review this one as well.

NPM-Behind-Proxy.md
```diff
@@ -0,0 +1,40 @@
+# NPM Behind A Proxy Server
+
+### Use-cases
+
+You might need to modify `npm install` commands that accesses remote repositories ([npmjs](https://www.npmjs.com/), for example) for installing Node JS modules; if your internet access is through a [proxy server](https://en.wikipedia.org/wiki/Proxy_server).
```
SeanFCC: * I would change 'accesses' to 'access'.

NPM-Behind-Proxy.md
```diff
@@ -0,0 +1,40 @@
...
+
+It's quite common if you are a student in college, or maybe you are a working professional; accessing the internet behind an institute/company mandated firewall.
```
SeanFCC: Maybe reword to something like: 'Proxy servers are common in college and business type environments.'

NPM-Behind-Proxy.md
```diff
@@ -0,0 +1,40 @@
...
+```
+You would need to replace `<username>`, `<password>`, `<proxy-server-url>`, `<port>` with the values specific to your proxy server credentials.
+
+These fields are optional. For instance, your proxy server might not even require `<username>` and `<password>`, or that it might be running on port 80 (in which case `<port>` is not required).
+
+Once you have set these, your `npm install`, `npm i -g` etc. would work properly.
+
+### When Not to Use
+
+You should not have to use `npm` commands with proxy settings, if either of the following happens
```
SeanFCC: Suggestion:  '...if either of the following happens:'

NPM-Behind-Proxy.md
```diff
@@ -0,0 +1,40 @@
...
+
+### Using Proxy with NPM
+
+Once you have obtained the proxy settings (server URL, port, username and password); you need to configure your `npm` configurations as follows.
+
+```bash
+$ npm config set proxy http://<username>:<password>@<proxy-server-url>:<port>
+$ npm config set https-proxy http://<username>:<password>@<proxy-server-url>:<port>
+```
+You would need to replace `<username>`, `<password>`, `<proxy-server-url>`, `<port>` with the values specific to your proxy server credentials.
```
SeanFCC: Suggested change: I would remove 'You would need to' and have the sentence start with 'Replace'.

NPM-Behind-Proxy.md
```diff
@@ -0,0 +1,40 @@
...
+
+These fields are optional. For instance, your proxy server might not even require `<username>` and `<password>`, or that it might be running on port 80 (in which case `<port>` is not required).
+
+Once you have set these, your `npm install`, `npm i -g` etc. would work properly.
```
SeanFCC: Suggested change: instead of 'would work properly' I would say it 'should'.

NPM-Behind-Proxy.md
```diff
@@ -0,0 +1,40 @@
...
+
+>- Your system administrator or corporate policy does not allow you to access remote `npm` repositories from NPM-JS, for instance.
+- The remote repository of Node modules in question is not in your machine, but it's within the internal network.
+
+### Unset Proxy Settings
+
+Use [this](http://luxiyalu.com/how-to-remove-all-npm-proxy-settings/) blog post to unset your proxy settings. You can also manually remove the lines specifying your proxy settings from your `.npmrc` [file](https://docs.npmjs.com/files/npmrc).
+
+### Resources
+You may use the following for further reading on this
```
SeanFCC: Suggested Change: 'You may use the following for further reading on this:'

PR #697, opened 2016-04-08
Change while loops in last two solutions to use only 3 assignments

Ramapitecus:
I just simplified the swap that is being made in the while loops of the last two example solutions. I think this way it's clearer.

### DIFF
--- a/Algorithm-Smallest-Common-Multiple.md
+++ b/Algorithm-Smallest-Common-Multiple.md
@@ -113,32 +113,48 @@ Note: The solution above requires over 2,000 loops to calculate the test case
 
 ```js
 function smallestCommons(arr) {
-  for (i = Math.max(...arr); i >= Math.min(...arr); i--) { arr.push(i); }
+    var range = [];
+    for (var i = Math.max(arr[0], arr[1]); i >= Math.min(arr[0], arr[1]); i--) {
+	range.push(i);
+    }
 
-  return arr.slice(2).reduce(function(x, y) {
-    var a = x, b = y, t = 0;
-    while (a % b) { a = a % b; t = a; a = b; b = t; }
-    return x / b * y;
-  });
+    return range.reduce(function(previousValue, currentValue) {
+	var gcdPrevCurr = gcd(previousValue, currentValue);
+	return (previousValue * currentValue) / gcdPrevCurr;
+    });
+
+    function gcd(x, y) {	// Implements The Euclidean Algorithm
+	if (y === 0)
+	    return x;
+	else
+	    return gcd(y, x%y);
+    }
 }
 ```
 
 If you have trouble understanding the `.reduce()` method, you can compare it to
 this code, which essentially does the same thing:
 
-```
+```js
 function smallestCommons(arr) {
-  for (i = Math.max(...arr); i >= Math.min(...arr); i--) { arr.push(i); }
+    var range = [];
+    for (var i = Math.max(arr[0], arr[1]); i >= Math.min(arr[0], arr[1]); i--) {
+	range.push(i);
+    }
 
-  var lcm = arr[2], a = 0, b = 0, t = 0;
+    var lcm = range[0];
+    for (i = 1; i < range.length; i++) {
+	var GCD = gcd(lcm, range[i]);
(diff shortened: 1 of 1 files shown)


Rafase282:
I tested it and it works, the question is whether it is a good practice or not.

Rafase282:
@abhisekp @raisedadead @alayek

alayek:
@Ramapitecus thank you for this article. I sure like your enthusiasm that you want to contribute to the FCC Wiki and improve on even the tiniest bit!

Yes, I agree that it is easier to read with three assignments - but it's also not a _drastic_ improvement in readability. 

I won't jump on the performance wagon too, because the code you write is not the code that runs - the JIT modifies and optimizes it heavily before invoking the engine. Those four assignments won't result in any tangibly poor performance difference as compared to three assignments.

As far as readability is concerned, you might be able to try to improve it my adding comments to suggest intention. But if someone can read one without any assistive comments, there is a good chance they can read the other too.

I am torn, but I am inclined to **NOT** merge this.

Ramapitecus:
@alayek.Thanks for your comments. I don't think it would be a big improvement in performance. But I think it is a small improvement in readability. I found it confusing at first view. A weird way of swaping two variables. While I wrote the swap as you see everywhere and I think it's immediately clear what you are doing. And I concur that a comment would also be appropriate. To clarify that after the loop, `b` contains the GCD. And mention that it's implementing Euclid's algorithm.

SaintPeter:
I'm not sure that I even understand the difference in the code.

I think this should be re-written entirely with longer than one letter variable names.  Unless they're indexes, single letter variable names are generally a bad practice.  It's almost impossible to tell what is going on.

If you want to add to the value of this article, that's what I suggest you do.

Ramapitecus:
I think the original author of the code wanted to be clever and do this in the least amount of lines as possible. that is why he used multiple assignments in one line. and even wrote the entire while loop. To make it clear, I would write that loop as an internal function, named `gcd`. That would make it's intention clear. If that is OK with you guys. And maybe change the variable names, and leave one assignment per line.

alayek:
@Ramapitecus irrespective of intentions of author of original code snippet; please try to add value by adding comments, or making variable names meaningful; like @SaintPeter mentioned.

Ramapitecus:
thank you all for you comments. I'll redo this and try to include all your suggestions.

Ramapitecus:
What would be the correct way to resubmit this? Should the pull request be closed? I would then, redo the branch, and make a single commit to it with all the changes? Should I Just make a new commit to it, and squash them?

Rafase282:
Make the necessary commits and then squash.

Ramapitecus:
I have refactored the code to extract the grand common divisor computation into a separate local function. Also, I changed some variable names to more descriptive ones. I also changed the `...arr` in the `for` loop, because it caused a warning when used inside the editor in the challenge page. Finally I added a missing `js` to the last code block. 

I hope this addresses you suggestions. Thanks for you help and comments.

Algorithm-Smallest-Common-Multiple.md
```diff
@@ -113,32 +113,46 @@ Note: The solution above requires over 2,000 loops to calculate the test case
...
-
-  var lcm = arr[2], a = 0, b = 0, t = 0;
-
-  for (i = 3; i < arr.length; i++) {
-    a = lcm, b = arr[i];
-    while (a % b) { a = a % b; t = a; a = b; b = t; }
-    lcm = lcm / b * arr[i];
-  }
+    var range = [];
+    for (i = Math.max(arr[0], arr[1]); i >= Math.min(arr[0], arr[1]); i--)
```
SaintPeter: Can we get curly braces for this?  They're a best practice for code uglifiers.

SaintPeter:
The code is looking a LOT more readable.

I'd be willing to merge as is.  Maybe we can come back with a second pass to add a more detailed explanation of what we're actually doing in the Alternate Solution.

alayek:
@Ramapitecus please fix the curly braces issue as @SaintPeter has mentioned and we are good to go!

# LGTM :+1:

Ramapitecus:
@alayek Done.

SaintPeter:
:+1:

PR #705, opened 2016-04-09
Removed Duplicate and Fixed HTML5 Issue

theflametrooper:
Can you please squash my commits because i'm editing this in the browser... @Rafase282


Campsites.md
```diff
@@ -6,11 +6,11 @@ Whether you live in a major metropolitan area, or a small town, we encourage you
 
 Here are some helpful links:
 
-[Complete Alphabetical List of Campsites](List-of-Free-Code-Camp-city-based-Campsites)
+[Complete Alphabetical List of Campsites](List-of-Free-Code-Camp-city-based-Campsites.md)
```
abhisekp: No need to append the extension `.md`

theflametrooper: It doesn't work, It cannot find the page because there are no file extensions

abhisekp:
### **Note to Giant Squasher**
Please **keep the commit** with commit message `Removed Duplicate Article` as a separate commit. And squash all other commits.

Rafase282:
@abhisekp this is what it would be if I squash:

* Updated to HTML5 em tag

* Removed Duplicate Article

* Fixed Links on Campsites.md

* Removed my mistakes

It will also be one commit. If I do it the old way I would have to create a new PR on the repo with the squashing the way you ask and have it automatically close this one.

abhisekp:
@Rafase282 Use diff tools and see that there is absolutely no change in the last 2 commits. Last is simply a revert of the last of one.

Keep only the first two i.e.
- Updated to HTML5 em tag
- Removed Dupe Article

the last two can be skipped *happily* while rebasing.

And yes, you've to do in the old style. :dancer: :wink:

Rafase282:
@theflametrooper I think you deleted the branch with the changes. I would suggest you open two new PR, one for the html fix and the other for removing the duplicate file. Try to keep unrelated changes separated.

theflametrooper:
@Rafase282 Sure

PR #706, opened 2016-04-09
Removed Irrelevant and Pointless Articles

theflametrooper:
Camper News has been discontinued. I feel the Bootstrap article is pointless. Maybe someone could make a longer article for bootstrap in the future.


abhisekp:
@theflametrooper I don't think we should remove the `Bootstrap.md` file as it can act as a placeholder/linking article for other bootstrap articles in the wiki. 

Better to file an issue to link other Bootstrap articles in it.

theflametrooper:
@abhisekp One of the links doesn't even work. I'm already working on a more in-depth Bootstrap Article.

Rafase282:
@theflametrooper once your file is ready we can merge this one. Although it would be better to just edit the current boostrap file unless youa re using a different name.

PR #717, opened 2016-04-11
Article: How to Stream Live Coding on FreeCodeCamp

theflametrooper:
I updated this page to the newest requirements of how we let users stream. Leave any suggestions below!


Rafase282:
Why remove the Twitch.md file? It even points out to the right file.

theflametrooper:
@Rafase282 I didn't see a point after having lots of other pages related to Twitch. If you want I can just update the Twitch.md to the newest requirements.

PR #745, opened 2016-04-17
Undo git add article

aaronTHJ:
Article to undo ```git add``` before committing.

### DIFF
--- a/Git-Undo-Redo-A-Commit.md
+++ b/Git-Undo-Redo-A-Commit.md
@@ -31,8 +31,16 @@ Or alternatively you can use caps `-C`, where git will skip the editor window, a
 
 Re-using the "Same" commit message is also known as redoing/recommiting.
 
+#### Unstage before a commit
+
+To undo a change staged before a commit simply run `git reset <file>` or `git reset` to unstage all changes before a commit.
+
+Note: In older versions of git, the commands were `git reset HEAD <file>` and `git reset HEAD` respectively. This was changed in Git 1.8.2 
+
 #### Some More tricks:
 
 You can go back any number of commits by using `git reset --soft HEAD~n` where you want to undo last `n` commits.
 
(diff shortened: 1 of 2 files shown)


Rafase282:
If you are doing inline code then you just need one backtick *`*

awesomeaniruddh:
This is a useful topic, but does it merit its own article?

Rafase282:
Also, filename should be `Git-Undo-Add.md`

awesomeaniruddh:
I just checked and we do have a very similar (although not exactly the same) article: https://www.freecodecamp.com/wiki/en/git-undo-redo-a-commit/

I think we can probably modify that article slightly @Rafase282?

Rafase282:
@awesomeaniruddh what do you suggest?

awesomeaniruddh:
@Rafase282 
I don't know really. The topic content itself is really useful and should be added to the wiki, but it's too small to be its own article in my opinion?

Modifying the article I linked to be more general and about both commits and staged files makes the most sense to me.

alayek:
@awesomeaniruddh what's the standard minimum size (in terms of line no. or word count) for a wiki article?

Rafase282:
Well originally according to [this](https://github.com/FreeCodeCamp/FreeCodeCamp/wiki/Wiki-Style-Guide) 

>Keep your wiki articles short and focused.

awesomeaniruddh:
Yeah well, it's just that I thought that this article was *too short*

abhisekp:
I don't think is anything difficult or tricky thing in git to do. And as @awesomeaniruddh says the article is *too short* to add. So it's better to add in other article.

alayek:
I was thinking maybe we add this with the git add article, instead of creating another standalone article.

Rafase282:
https://github.com/aaronTHJ/wiki/pull/2
https://github.com/aaronTHJ/wiki/pull/1

Issue #753, opened 2016-04-19
max-width to .wiki-container


xRahul:
`max-width` css to the class `.wiki-container` is set to `960px` which is making the UI too narrow.
Need to remove that.

Without `max-width: 960px;`

![screenshot from 2016-04-19 10 57 07](https://cloud.githubusercontent.com/assets/1639945/14629548/22c94eee-0625-11e6-8472-631d18a9fa34.png)

With `max-width: 960px;`

![screenshot from 2016-04-19 10 56 31](https://cloud.githubusercontent.com/assets/1639945/14629549/22c9bafa-0625-11e6-9b61-ad4c8820784f.png)

alayek:
@xRahul if `max-width: 960px` is removed, are you able to properly navigate horizontally to view the full image?

Also, this issue should be filed on the main repo: https://github.com/FreeCodeCamp/FreeCodeCamp because it's a site level presentation issue, not an issue with the content.

xRahul:
@alayek Can you give an example of which image you're talking about?
I just checked on the wiki, and we can still do the horizontal scrolling after removing the max-width.

And this is an issue with the gh-pages branch of the wiki, not the main repo.

raisedadead:
@xRahul Please log this on the main repo.

xRahul:
@raisedadead @alayek  this is an issue with the `gh-pages` branch of wiki repo itself. It doesn't have anything to do with the main repo.
file: https://github.com/FreeCodeCamp/wiki/blob/gh-pages/styles.css

raisedadead:
@xRahul Thanks. 

Its not about the location of the code. @alayek, myself couldn't be more correct in saying where the issue needs to be logged, considering that's our primary responsibility.

FYI, I'll give you how I reached the conclusion of pointing you to the main repo. Check [these issues](https://github.com/FreeCodeCamp/FreeCodeCamp/labels/wiki).

The `gh-pages` is supposed to be deployed via scripts via [this generator](https://github.com/FreeCodeCamp/wiki-generator), which is later embedded in an iframe in at `/wiki` on the main website.

Berkeley did some quick routing and CSS fixes since gatsby that the wiki-generator depends upon is now under works. You are more than welcome to take a look at the generator and come up with fixes.

As sure we need all the help we can get, but its not cool to dandy in and just say:
> this is an issue with the gh-pages branch of wiki repo itself. It doesn't have anything to do with the main repo.

The issues are triage-ed for valid reasons.

[...]

PR #755, opened 2016-04-19
Update styles.css

xRahul:
**Description**
Removed max-width:960px from .wiki-container

**File**
https://github.com/freecodecamp/wiki/blob/gh-pages/styles.css

**_URL to the panel_**
https://www.freecodecamp.com/wiki/en/

**Before removing the CSS**
screenshot from 2016-04-19 10 56 31

**After Removing the CSS**
screenshot from 2016-04-19 10 57 07


xRahul:
Related issue:
https://github.com/FreeCodeCamp/wiki/issues/753

alayek:
@xRahul thanks for this PR. But, even though it's a small change in CSS; it's still a site wide change that could affect user experience of a thousands of users.

Please don't mind, I have to block the PR for now. It requires a discussion and quorum between core team members. We will get back to this as soon as possible.

In the meantime, you can chat with us in the [Wiki room](https://gitter.im/FreeCodeCamp/Wiki)

xRahul:
@alayek sure! I get that the change is not conventional and being a change in the core css of the wiki site, it will require some discussion and testing beforehand.

abhisekp:
@xRahul you need not close this though. Let it keep open for review and then if approved, this can be merged.

xRahul:
alright! The change is small, so we can make them after discussion.

alayek:
Nonetheless, the PR should remain open. The `blocked` label makes sure no mod would merge it unless the original label-er removes it.

raisedadead:
@hallaathrad I know you are super busy, but @xRahul has done some good work over here, would you please take a look.

Warning:
This may need to be taken up to the wiki-generator repo when that goes into works.

Rafase282:
Have you tested the responsiveness?

hallaathrad:
Hey all! Thanks go to @xRahul for his contribution!
I'm ok with the change, except it doesn't really fix the underlying issue (what if the screen is actually not broader than 1024px? the images are not adjusting to the screen size so they'd still be hidden for smaller resolutions... for instance). Besides, there's a design principle for lines of text not being longer than 13 words' length, or paragraphs start looking wonky and legibility decreases.

Furthermore, and this is on the procedural side of things. This PR should not be against gh-pages for this one gets overwritten by the generator... @xRahul chat me up on gitter and we can get this thing worked out properly, I'll help if you want me to ;)

It was in my ToDos anyway. What say you?

xRahul:
@hallaathrad Thanks man for making it clear. 
Wiki-generator seems awesome, I'll check it out for any future changes.

PR #757, opened 2016-04-19
Added Sans Serif Font

theflametrooper:
**_Before:_**

**_After:_**

### DIFF
--- a/404.html
+++ b/404.html
@@ -5,4 +5,4 @@
 <br>
 <br>
 <h1 style='text-align: center;'>404: There is no article at this address.</h1>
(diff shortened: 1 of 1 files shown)


Rafase282:
Slightly better but personally I don't see much improvement with the new font. However, with regards to the PR, :+1:

raisedadead:
How about https://www.google.com/fonts/specimen/Raleway ?
Light 300 should be good IMO

theflametrooper:
@raisedadead Let me try and i'll send you a screenshot of what it looks locally...

PR #762, opened 2016-04-20
Added Raleway Font

theflametrooper:
#### The anchor tags may need updated. What do you think @raisedadead ?
---------------
_Screenshot of new design:_


raisedadead:
LGTM. But as always be aware that the gh-pages will get overwritten by wiki-generator. So it will have to be taken up then once again.

Rafase282:
Why not directly modify the files so the generator will keep the style changes?

theflametrooper:
@Rafase282 Where is that file located? I was unaware of it's existance...

theflametrooper:
Closing this pr and will open a new one here: https://github.com/FreeCodeCamp/wiki-generator

PR #773, opened 2016-04-21
Add Java Inheritance wiki

koustuvsinha:
Added Inheritance article as per #534. Please have a look @alayek , @Rafase282, @abhisekp

### DIFF
--- a/Java-Inheritance.md
+++ b/Java-Inheritance.md
@@ -163,7 +163,7 @@ public class Vehicle() {
 
 public class Car extends Vehicle {
 	public Car() {
-		super();
+	  super();
 	}
 	public void run() {
 	  super.start();
@@ -171,6 +171,8 @@ public class Car extends Vehicle {
 }
 ```
 
-Remember, if the superclass does not have any constructors defined, you dont have to call it explicitely in the subclass. Java handles that internally for you! Also, invocation to `super` constructor is done in the case when the super class is to be called with any other constructor other than the default constructor.
+Remember, if the superclass does not have any constructors defined, you dont have to call it explicitely in the subclass. Java handles that internally for you! Invocation to `super` constructor is done in the case when the super class is to be called with any other constructor other than the _default constructor_.
+
+If no other constructors are defined, then Java invokes the default super class constructor (*even if not defined explicitly*).
 
 Congrats, now you know all about Inheritance! Read more about advanced ways to inherit things in [Abstract Classes]() and [Interfaces]()!
--- a/Java-Inheritance.md
+++ b/Java-Inheritance.md
@@ -1,0 +1,168 @@
+# Java Inheritance
+
+Java inheritance refers to the ability of a Java Class to _inherit_ the properties from some other Class. Think of it like a child inheriting properties from its parents, the concept is very similar to that. In Java lingo, it is also called _extend_-ing a class. Some simple things to remember :
+
+* The Class that extends or inherits is called **subclass**
+* The Class that is being extended or inherited is called **superclass**
+
+Thus, inheritance gives Java the cool capability of _re-suing_ a code, or share codes between classes! Lets describe it with the classic example of a `Vehicle` class and a `Car` class :
+
+```java
+public class Vehicle {
+	public void start() {
+		// starting the engine
+	}
+
+	public void stop() {
+		// stopping the engine
+	}
+}
+
+public class Car extends Vehicle {
+	int numberOfSeats = 4;
+
+	public int getNumberOfSeats() {
+		return numberOfSeats;
+	}
+}
+```
+
+Thus, here we can see the `Car` class inheriting the properties of the `Vehicle` class. So, we dont have to write the same code of `start()` and `stop()` for Car as well, as those properties comes from its parent. Yes, objects created from the `Car` class _also_ has those properties!
+
+```java
+Car tesla = new Car();
+
+tesla.start();
+
+tesla.stop();
+```
+
+But, does the parent class has the methods of the child? No, it doesn't.
+
+Thus, whenever you need to share some common bit of code to some more classes, it is always good to have a parent Class, and then extend that Class whenever needed! Saves a lot of lines of code, makes code modular and better testable.
+
+### What can be inherited ?
+
+* All `protected` and `public` fields and methods from parent
+
+### What cannot be inherited ?
+
+* `private` fields and methods
+* Constructors. Although, subclass constructor _has_ to call superclass constructor (More on that later!)
+* Multiple classes. Java supports only Single Inheritance, that is you can only inherit one class at a time.
+* Fields. Individual fields of a class cannot be overriden by the subclass.
+
+## Type Casting & Reference
+
+In Java, it is possible to reference a subclass as an _instance_ of its superclass. For example, `Car` class object can be referenced as a `Vehicle` class instance like this : 
+
+```java
+Vehicle car = new Car();
+```
+
+Although, the opposite is not possible :
+
+```java
+Car car = new Vehicle(); // ERROR
+```
+
+So, since you can reference a Java subclass as a superclass instance, you can cast an subclass object easily to a superclass instance. It may be possible to cast a superclass object into a subclass type, but _only if the object is really an instance of subclass_. So keep this in mind :
+
+```java
+Car car = new Car();
+Vehicle vehicle = car; // upcasting
+Car car2 = (Car)vechile; //downcasting
+
+Bike bike = new Bike(); // say Bike is also a subclass of Vehicle
+Vehicle v = bike; // upcasting, no problem here.
+Car car3 = (Car)bike; // Compilation Error : as bike is NOT a instance of Car
+```
+
+Now you know how to share code through parent - child relationship. But, what if, you do not like the implementation of a particular method in the child class and want to write a new one for it? What to do then?
+
+## Override it!
+
+Java lets you override or redefine the methods defined in the superclass. For example, your `Car` class has a different implementation of `start()` than the parent `Vehicle`, so you do this :
+
+```java
+public class Vehicle() {
+	public void start() {
+	  System.out.println("Vehicle start code");
+	}
+}
+
+public class Car extends Vehicle {
+	public void start() {
+	  System.out.println("Car start code");
+  }
+}
+
+Car car = new Car();
+car.start(); // "Car start code"
+```
+
+So, its pretty simple to override methods in the subclass. Although, there is a _catch_. Only that superclass method will be overriden which has the _exact same method signature_ as the subclass method. Thus, `public void start(String key)` would not override `public void start()`.
+
+**Notes** :
+
+* You cannot override private methods of the superclass. (Quite obvious, isn't it?)
+* What if the method of superclass which you are overriding in the subclass suddenly gets obliterated or methods changed? It would fail in runtime! So Java provides you a nifty annotation `@Override` which you can place over the subclass method, which will warn the compiler of those incidents!
+
+## How to call super class methods?
+
+Funny you ask about it! Just use the keyword `super` :
+
+```java
+public class Vehicle() {
+	public void start() {
+	  System.out.println("Vehicle start code");
+	}
+}
+
+public class Car extends Vehicle {
+	public void run() {
+	  super.start();
+  }
+}
+
+Car car = new Car();
+car.run(); // "Vehicle start code"
+```
+
+## How to know the type of a class?
+
+Using the `instanceof` keyword. Having lots of classes and subclasses it would be a little confusing to know which class is a subclass of which one in runtime. So, we can use `instanceof` to determine whether a class is actually a subclass of a superclass or not.
+
+```java
+Car car = new Car();
+
+boolean flag = car instanceof Vehicle; // true in this case!
+```
+
+## Constructors & Inheritance
+
+As mentioned earlier, constructors cannot be directly inherited by a subclass. Although, a subclass is _required_ to call its parent's constructor as the first thing in its own constructor. How? You guessed it, using `super` :
+
+```java
+public class Vehicle() {
+	public Vehicle() {
+		// constructor
+	}
+	public void start() {
+	  System.out.println("Vehicle start code");
+	}
+}
+
+public class Car extends Vehicle {
+	public Car() {
+		super();
+	}
+	public void run() {
+	  super.start();
+  }
+}
+```
+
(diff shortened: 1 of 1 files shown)


Java-Inheritance.md
```diff
@@ -0,0 +1,168 @@
+# Java Inheritance
+
+Java inheritance refers to the ability of a Java Class to _inherit_ the properties from some other Class. Think of it like a child inheriting properties from its parents, the concept is very similar to that. In Java lingo, it is also called _extend_-ing a class. Some simple things to remember :
```
alayek: >  Java ~~Class~~ `Class` to ~~_inherit_~~ `inherit`

Java-Inheritance.md
```diff
@@ -0,0 +1,168 @@
...
+
+* The Class that extends or inherits is called **subclass**
+* The Class that is being extended or inherited is called **superclass**
+
+Thus, inheritance gives Java the cool capability of _re-suing_ a code, or share codes between classes! Lets describe it with the classic example of a `Vehicle` class and a `Car` class :
```
alayek: > capability of ~~_re-suing_~~ _re-using_ a code

alayek: > share ~~codes~~ code between classes

alayek: > ~~Lets~~ Let's describe it with the

Java-Inheritance.md
```diff
@@ -0,0 +1,168 @@
...
+public class Car extends Vehicle {
+	int numberOfSeats = 4;
+
+	public int getNumberOfSeats() {
+		return numberOfSeats;
+	}
+}
+```
+
+Thus, here we can see the `Car` class inheriting the properties of the `Vehicle` class. So, we dont have to write the same code of `start()` and `stop()` for Car as well, as those properties comes from its parent. Yes, objects created from the `Car` class _also_ has those properties!
```
alayek: > ~~Thus, here~~Here we can see the `Car`

alayek: > So, we ~~dont~~ don't

alayek: > for ~~Car~~ `Car` as well,

alayek: > as those properties ~~comes~~ come from

alayek: >---
```diff
- Yes, objects created from the `Car` class _also_ has those properties!
+ Yes, objects created from the `Car` class _also_ have those properties!
```

Java-Inheritance.md
```diff
@@ -0,0 +1,168 @@
...
+
+```java
+Car tesla = new Car();
+
+tesla.start();
+
+tesla.stop();
+```
+
+But, does the parent class has the methods of the child? No, it doesn't.
```
alayek: > does the parent ~~class~~ Class

Java-Inheritance.md
```diff
@@ -0,0 +1,168 @@
...
+
+Thus, whenever you need to share some common bit of code to some more classes, it is always good to have a parent Class, and then extend that Class whenever needed! Saves a lot of lines of code, makes code modular and better testable.
```
alayek: Too much _Thus_ I think. Use _Therefore_ or _Hence_.

Java-Inheritance.md
```diff
@@ -0,0 +1,168 @@
...
+
+### What can be inherited ?
+
+* All `protected` and `public` fields and methods from parent
+
+### What cannot be inherited ?
+
+* `private` fields and methods
+* Constructors. Although, subclass constructor _has_ to call superclass constructor (More on that later!)
```
alayek: > Although, subclass constructor _has_ to call superclass constructor

That's not always true, is it?

Java-Inheritance.md
```diff
@@ -0,0 +1,168 @@
...
+* Multiple classes. Java supports only Single Inheritance, that is you can only inherit one class at a time.
```
alayek: > Java supports only ~~Single Inheritance~~ **single inheritance**

Highlight the part about single inheritance.

Java-Inheritance.md
```diff
@@ -0,0 +1,168 @@
...
+* Fields. Individual fields of a class cannot be overriden by the subclass.
+
+## Type Casting & Reference
+
+In Java, it is possible to reference a subclass as an _instance_ of its superclass. For example, `Car` class object can be referenced as a `Vehicle` class instance like this :
```
alayek: Clearly mention this is an implementation of the `Polymorphism` concept from OOP.

Java-Inheritance.md
```diff
@@ -0,0 +1,168 @@
...
+Vehicle car = new Car();
+```
+
+Although, the opposite is not possible :
+
+```java
+Car car = new Vehicle(); // ERROR
+```
+
+So, since you can reference a Java subclass as a superclass instance, you can cast an subclass object easily to a superclass instance. It may be possible to cast a superclass object into a subclass type, but _only if the object is really an instance of subclass_. So keep this in mind :
```
alayek: > So, since you can reference a Java subclass as a

No need to use _since_ followed by a _So_. Use anyone of the them. But not both in consecution.

alayek: > you can cast ~~an~~ a subclass object

Java-Inheritance.md
```diff
@@ -0,0 +1,168 @@
...
+Car car = new Car();
+Vehicle vehicle = car; // upcasting
+Car car2 = (Car)vechile; //downcasting
+
+Bike bike = new Bike(); // say Bike is also a subclass of Vehicle
+Vehicle v = bike; // upcasting, no problem here.
+Car car3 = (Car)bike; // Compilation Error : as bike is NOT a instance of Car
+```
+
+Now you know how to share code through parent - child relationship. But, what if, you do not like the implementation of a particular method in the child class and want to write a new one for it? What to do then?
```
alayek: > ~~parent - child~~ parent-child relationship

Java-Inheritance.md
```diff
@@ -0,0 +1,168 @@
...
+
+## Override it!
+
+Java lets you override or redefine the methods defined in the superclass. For example, your `Car` class has a different implementation of `start()` than the parent `Vehicle`, so you do this :
```
alayek: > Java lets you ~~override~~ `override`

Since "override" is an actual term, better inline-code it. Or italicize. Since you are using it the first time.

Java-Inheritance.md
```diff
@@ -0,0 +1,168 @@
...
+	public void start() {
+	  System.out.println("Car start code");
+  }
+}
+
+Car car = new Car();
+car.start(); // "Car start code"
+```
+
+So, its pretty simple to override methods in the subclass. Although, there is a _catch_. Only that superclass method will be overriden which has the _exact same method signature_ as the subclass method. Thus, `public void start(String key)` would not override `public void start()`.
```
alayek: > So, ~~its~~ it's pretty

alayek: Add one more line here, about what signature means. All arguments to the method in same order and of same type.

Java-Inheritance.md
```diff
@@ -0,0 +1,168 @@
...
+
+**Notes** :
+
+* You cannot override private methods of the superclass. (Quite obvious, isn't it?)
+* What if the method of superclass which you are overriding in the subclass suddenly gets obliterated or methods changed? It would fail in runtime! So Java provides you a nifty annotation `@Override` which you can place over the subclass method, which will warn the compiler of those incidents!
```
alayek: Highlight the fact that that annontations like that in Java are good coding practice, not a necessity. The compiler is still smart enough to figure out overriding on its own.

Emphasize that unlike other OOP languages, like Python or TypeScript, it doesn't necessarily modify the method or add extra functionality.

Java-Inheritance.md
```diff
@@ -0,0 +1,168 @@
...
+public class Car extends Vehicle {
+	public void run() {
+	  super.start();
+  }
+}
+
+Car car = new Car();
+car.run(); // "Vehicle start code"
+```
+
```
alayek: Add a remark here that though you can go up one level of `super` call, you cannot go up the inheritance chain with nested `super` calls.

Java-Inheritance.md
```diff
@@ -0,0 +1,168 @@
...
+```java
+Car car = new Car();
+
+boolean flag = car instanceof Vehicle; // true in this case!
+```
+
+## Constructors & Inheritance
+
+As mentioned earlier, constructors cannot be directly inherited by a subclass. Although, a subclass is _required_ to call its parent's constructor as the first thing in its own constructor. How? You guessed it, using `super` :
```
alayek: `super` constructor is not always the first thing it's supposed to call. It can call that after initializing its own instance variables.

Java-Inheritance.md
```diff
@@ -0,0 +1,168 @@
...
+	}
+	public void run() {
+	  super.start();
+  }
+}
+```
+
+Remember, if the superclass does not have any constructors defined, you dont have to call it explicitely in the subclass. Java handles that internally for you!
+
+Congrats, now you know all about Inheritance! Read more about advanced ways to inherit things in [Abstract Classes]() and [Interfaces]()!
```
alayek: @abhisekp take a look at this line. Should these links be blank, because the articles are not yet added, but would be added soon.

alayek:
@theflametrooper might wanna take a look for typos and all. Technically, this article is correct and code snippets work the way they are supposed to, as described in the article.

@koustuvsinha can't thank you enough!

Java-Inheritance.md
koustuvsinha: Is it? http://stackoverflow.com/questions/1168345/why-does-this-and-super-have-to-be-the-first-statement-in-a-constructor

Java-Inheritance.md
```diff
@@ -0,0 +1,172 @@
+# Java Inheritance
+
+Java inheritance refers to the ability of a Java Class to `inherit` the properties from some other Class. Think of it like a child inheriting properties from its parents, the concept is very similar to that. In Java lingo, it is also called _extend_-ing a class. Some simple things to remember :
+
+* The Class that extends or inherits is called **subclass**
+* The Class that is being extended or inherited is called **superclass**
+
+Thus, inheritance gives Java the cool capability of _re-using_ a code, or share code between classes! Let's describe it with the classic example of a `Vehicle` class and a `Car` class :
```
theflametrooper: ~~_re-using_ a code~~ _re-using_ code

Java-Inheritance.md
alayek: Ahh..ok. In that case, link to it like this:

> call its parent's constructor as the [first thing](http://stackoverflow.com/questions/1168345/why-does-this-and-super-have-to-be-the-first-statement-in-a-constructor) in its own constructor.

koustuvsinha:
@alayek @theflametrooper updated the typos & implemented suggestions

Java-Inheritance.md
abhisekp: yes. :+1:

Java-Inheritance.md
```diff
@@ -0,0 +1,172 @@
...
+		// constructor
+	}
+	public void start() {
+	  System.out.println("Vehicle start code");
+	}
+}
+
+public class Car extends Vehicle {
+	public Car() {
+		super();
```
abhisekp: Invocation to `super` constructor is done in the case when the super class is to be called with any other constructor other than the default constructor.

Java-Inheritance.md
```diff
@@ -0,0 +1,172 @@
...
+	}
+	public void run() {
+	  super.start();
+  }
+}
+```
+
+Remember, if the superclass does not have any constructors defined, you dont have to call it explicitely in the subclass. Java handles that internally for you!
```
abhisekp: Also, invocation to `super` constructor is done in the case when the super class is to be called with any other constructor other than the default constructor.

Java-Inheritance.md
```diff
@@ -0,0 +1,172 @@
+# Java Inheritance
+
+Java inheritance refers to the ability of a Java Class to `inherit` the properties from some other Class. Think of it like a child inheriting properties from its parents, the concept is very similar to that. In Java lingo, it is also called _extend_-ing a class. Some simple things to remember :
+
+* The Class that extends or inherits is called **subclass**
+* The Class that is being extended or inherited is called **superclass**
+
+Thus, inheritance gives Java the cool capability of _re-using_ code, or share code between classes! Let's describe it with the classic example of a `Vehicle` class and a `Car` class :
+
```
abhisekp: add a heading 2 break point here for our sweet beloved @camperbot to extract upto this point for gitter chat 

Like this,

```md
## (just add a space after the two ##)
```

Move the line `Let's describe it with the classic example of a `Vehicle` class and a `Car` class :` after the heading.

Java-Inheritance.md
```diff
@@ -0,0 +1,176 @@
...
+		// constructor
+	}
+	public void start() {
+	  System.out.println("Vehicle start code");
+	}
+}
+
+public class Car extends Vehicle {
+	public Car() {
+		super();
```
abhisekp: why the extra indentation?

Java-Inheritance.md
```diff
@@ -0,0 +1,176 @@
...
+	}
+	public void run() {
+	  super.start();
+  }
+}
+```
+
+Remember, if the superclass does not have any constructors defined, you dont have to call it explicitely in the subclass. Java handles that internally for you! Also, invocation to `super` constructor is done in the case when the super class is to be called with any other constructor other than the default constructor.
```
abhisekp: Invocation to `super` constructor is done in the case when the super class is to be called with any other constructor other than the default constructor. If no other constructors are defined, then Java invokes the default super class constructor (*even if not defined explicitly*).

alayek:
# LGTM :+1:

Java-Inheritance.md
```diff
@@ -0,0 +1,178 @@
...
+	  super.start();
+  }
+}
+```
+
+Remember, if the superclass does not have any constructors defined, you dont have to call it explicitely in the subclass. Java handles that internally for you! Invocation to `super` constructor is done in the case when the super class is to be called with any other constructor other than the _default constructor_.
+
+If no other constructors are defined, then Java invokes the default super class constructor (*even if not defined explicitly*).
+
+Congrats, now you know all about Inheritance! Read more about advanced ways to inherit things in [Abstract Classes]() and [Interfaces]()!
```
Rafase282: Are these placeholder links? Loos that way, just want to confirm. If they are, you should add the relative link that you think you would be using. When merged it will just try to create the new article but it will help you remember what to name them or give others an idea of what to name it.

PR #775, opened 2016-04-22
Addition to Algorithm-Search-and-Replace.md

Scoodood:
Algorithm-Search-and-Replace.md

Added code solution 4, the code is shorter and cleaner and yet still get the
job done efficiently.

PR #776, opened 2016-04-22
Added code solution 5 to algorithm-pig-latin

Scoodood:
Algorithm-Pig-Latin.md

Added code solution 5
Use regex group to solve the problem in just 3 lines

### DIFF
--- a/Algorithm-Pig-Latin.md
+++ b/Algorithm-Pig-Latin.md
@@ -108,6 +108,22 @@ function translatePigLatin(str) {
 }
 ```
 
+#### Fifth Solution (regex grouping)
+```js
+function translatePigLatin(str) {
+  // vowel = AEIOU  consonant = non-vowel;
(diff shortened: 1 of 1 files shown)


alayek:
@Scoodood thanks for the PR. 

You are using the RegEx output `$1`, `$2` to form the output, and none of the other four solutions do this. I think I saw some Ruby solution that used a similar trick in some tutorial, and I actually liked that!

----
I am willing to merge this. But I would let other mods take a look before that, because we have a considerably higher bar for code solution entry in our Wiki.

@abhisekp @SaintPeter @Rafase282 you might want to take a look and share your thoughts.

Rafase282:
@Scoodood please take a look at the discussion [here](https://github.com/FreeCodeCamp/wiki/pull/775)

This looks like it would be a great implementation to add, but not yet.

PR #779, opened 2016-04-23
Added Simpler For-In Loop Solution

Camoen:
This is the solution that popped out at me when I worked through the problem.  It seems similar to the Array.reduce() solution, just with a for-in loop rather than using .reduce().


Rafase282:
We are not accepting new solutions unless they new [template](https://github.com/FreeCodeCamp/FreeCodeCamp/wiki/Wiki-Template-Challenge-Solution) has been applied, please read this :point_right: [April 22, 2016 1:50 PM](https://gitter.im/FreeCodeCamp/Wiki?at=571a64759689a5440f7aa1c8)

You can also check a similar PR [here](https://github.com/FreeCodeCamp/wiki/pull/775)

PR #782, opened 2016-04-23
Update Checkpoint-Profile-Lookup.md

### LINKED ISSUE #778: Bad solution code for the "Profile lookup" challenge
link: https://github.com/FreeCodeCamp/FreeCodeCamp/wiki/Checkpoint-Profile-Lookup#code-solution

The code solution in the profile lookup challenge wiki page is wrong(I think), I added the correct code below with an explanation inside comments in front of the modified line.
```js
for (var x = 0; x < contacts.length; x++){
    if (contacts[x].firstName === firstName) {
        if (contacts[x].hasOwnProperty(prop)) {  //instead of contacts[x][prop] because in that case it would return the value of that prop and if that prop was e.g 0  the if statement wouldn't work has expected.
            return contacts[x][prop];
        } else {
            return "No such property";
        }
    }
}
return "No such contact";

```

X-43A:
# FreeCodeCamp - Wiki Pull Request Template

`.hasOwnProperty()` method is more appropriate in the code solution, I also updated the explanation.

fixes https://github.com/FreeCodeCamp/wiki/issues/778

### DIFF
--- a/Checkpoint-Profile-Lookup.md
+++ b/Checkpoint-Profile-Lookup.md
@@ -49,7 +49,7 @@ Remember to use [ Read-Search-Ask](How-to-get-help-when-you-get-stuck) if you ge
 ```
 for (var x = 0; x < contacts.length; x++){
     if (contacts[x].firstName === firstName) {
-        if (contacts[x][prop]) {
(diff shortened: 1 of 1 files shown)


n00bax:
closes #778

PR #783, opened 2016-04-24
Add today's cities, link to assets repo and improve the map's description

hallaathrad:
Just so that people stop looking for their campsites in the map before looking at the list (because who cares for "find", right?)


Rafase282:
Good, I don't know why people would use the map instead of `ctrl + f` but now they are warned about the outdated map.

hallaathrad:
at least once a week they make the request... The human mind blows mine sometimes.

PR #799, opened 2016-04-27
Article: Java Math.sqrt method

nag786:
This is reference to Math square root function in java

### DIFF
--- a/Math-sqrt-function.md
+++ b/Math-sqrt-function.md
@@ -1,0 +1,34 @@
+# Math.sqrt() Function in Java
+**Description:** Math.sqrt() method returns the square root value of the primitive data type .If a number is less than zero, it will return NaN.
+## Syntax:
+     double sqrt(double d)
+
+## Parameters:
+	Contains primitive data type
+## Return Value:
+	Returns the square root of argument.
+```Example:
+public class Math{
+    
+    public static void main(String[] args) {
+        
+        double a = Math.sqrt(9);//calculates square root of 9
+                
+        double b = Math.sqrt(2456.7);//calculates square root of 2456.7
+        
+        double c = Math.sqrt(-4568);//calculates square root of -4568
+        
+        System.out.println("Math Square root value :"+a);
+        System.out.println("Math Square root value :"+b);
+        System.out.println("Math Square root value :"+c);
+    
+}
+```
+
+```
+Output:
+Math Square root value:3.0
+Math Square root value :49.565108695532984
+Math Square root value :NaN
+```
+ Source: [Math.sqrt()](http://www.tutorialspoint.com/java/number_sqrt.htm)


abhisekp:
**File rename**:-
```diff
- Math-sqrt-function.md
+ Java-Math.sqrt.md
```

Math-sqrt-function.md
```diff
@@ -0,0 +1,34 @@
+# Math.sqrt() Function in Java
+**Description:** Math.sqrt() method returns the square root value of the primitive data type .If a number is less than zero, it will return NaN.
```
alayek:
``` diff
- Math.sqrt() method returns the square root value of the primitive data type .If a number is less than zero, it will return NaN.
+ `Math.sqrt()` method returns the square root value of the primitive data type .If a number is less than zero, it will return `NaN`.
```

Math-sqrt-function.md
```diff
@@ -0,0 +1,34 @@
...
+## Syntax:
+     double sqrt(double d)
```
alayek: Use multiline code syntax for Markdown

```java
double sqrt(double d)
```

And I think you mean signature. Not syntax.

Math-sqrt-function.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Parameters:
+	Contains primitive data type
+## Return Value:
+	Returns the square root of argument.
+```Example:
```
alayek:
```diff
- ```Example:
+ ```java
```

alayek: Write the heading _Example_ in a separate line.

Math-sqrt-function.md
```diff
@@ -0,0 +1,34 @@
...
+    
+}
+```
+
+```
+Output:
+Math Square root value:3.0
+Math Square root value :49.565108695532984
+Math Square root value :NaN
+```
```
alayek: Add a [REPL](https://repl.it/) code snippet here, so that anyone can execute the above code in a click directly within browser.

Math-sqrt-function.md
```diff
@@ -0,0 +1,34 @@
...
+ Source: [Math.sqrt()](http://www.tutorialspoint.com/java/number_sqrt.htm)
```
alayek: No, this is absolutely abominable. You cannot link to anything else other than official documentation hosted by Oracle (previously hosted by Sun).

If you want to include the above reference, include it under further reading or maybe as an extra reference. But the first reference must be the official doc.

alayek:
Thanks for the article @nag786 ! I have left some comments, please try to work on those. If you have any questions, do ask for help [here](https://gitter.im/FreeCodeCamp/Wiki).

@koustuvsinha might wanna take a look.

theflametrooper:
I'm gonna merge this article and then make a pr with the suggestions.

Issue #707, opened 2016-04-09
Create Articles: CSS Frameworks


theflametrooper:
This will include articles about popular css frameworks and the basics of how to use them.
The articles should include the following css frameworks:

- [x] [Bootstrap](http://getbootstrap.com/)
- [x] [Materialize](http://materializecss.com/)
- [x] [Foundation](http://foundation.zurb.com/)
- [x] [Skeleton](http://getskeleton.com/)
- [x] [Bulma](http://bulma.io/)

CEREBR4L:
They could almost do with their own one each with learning resources, cheat sheets and w/e else is needed. Bulma (a flexbox framework) might be a good one to do also?

theflametrooper:
@CEREBR4L Yeh, we'll create a separate article for each ;) Thanks, I added Bulma.

PR #802, opened 2016-04-27
Article: Materialize

theflametrooper:
Refer #707

### DIFF
--- a/Materialize.md
+++ b/Materialize.md
@@ -1,0 +1,32 @@
+# Materialize
+
+Materialize is a Responsive CSS framework based on Google's [Material Design](https://design.google.com/spec/) Language.
+
+### Getting Started
+
+Here is a simple HTML template which includes the latest compiled and minified CSS and Javascript for the Materialize library.
+We have used a CDN in this example, but you can checkout other ways of installing Materialize [here](http://materializecss.com/getting-started).
+
+```html
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <title>Materialize Template</title>
+    <link href="http://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet">
+    <link type="text/css" rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/materialize/0.97.6/css/materialize.min.css" media="screen,projection"/>
+  </head>
+  <body>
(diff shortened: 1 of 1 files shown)


Materialize.md
```diff
@@ -0,0 +1,33 @@
+# Materialize
+
+Materialize is a Responsive CSS framework based on Google's [Material Design](https://design.google.com/spec/) Language.
+
+### Getting Started
```
abhisekp: Use header 2

```md
- ### Getting Started
+ ## Getting Started
```

abhisekp:
@theflametrooper would be nice to add a mini tutorial for materialize.

Materialize.md
theflametrooper: @abhisekp Is this for all subheadings?

Rafase282: @theflametrooper normally you would go in ascending order, Heading 1, 2,3 .. n

The bigger the number the deeper it is into the root header. An article should ideally shoudl have One head 1, and the rest top headings would be heading 2, with subheadings as needed.

Rafase282:
Mini tutorial? A link to their docs would be enough in my opinion. I think these articles should be more about letting campers know about something other than pure css and boostrap. Later we might add info from their docs but I don't see how you would make a mini guide for frameworks.

abhisekp:
@theflametrooper @Rafase282 just adding a few relevant and important parts of the framework show casing the usefulness of the framework over pure CSS would be really beneficial. 
This will make the reader get attracted towards materialize (and CSS frameworks overall) and also let them get started with a very basic use case. 

_It really helps to get a kick start._ :boot: :8ball:

theflametrooper:
@abhisekp @Rafase282 I will add this in a later PR. For now, let's get this merged!

theflametrooper:
@alayek When you merge, please squash commits as I did most of this in browser ;)

Materialize.md
```diff
@@ -0,0 +1,33 @@
+# Materialize
+
+Materialize is a Responsive CSS framework based on Google's [Material Design](https://design.google.com/spec/) Language.
+
+## Getting Started
+
+Here is a simple HTML template which includes the latest compiled and minified CSS and Javascript for the Materialize library.
+We have used a CDN in this example, but you can checkout other ways of installing Materialize [here](http://materializecss.com/getting-started).
```
alayek: The line above can be below this, just above the example.

alayek:
# LGTM :+1: 

@theflametrooper please update as mentioned in line note.

PR #806, opened 2016-04-27
Split Python Basics Part 1

Rafase282:
First part of the spiting process, more to come.


alayek:
# LGTM :+1: 

@koustuvsinha do we need REPL snippets? Also, perhaps we can interlink them with each other, other than linking them all to confluence type one single parent document?

theflametrooper:
## LGTM :+1:

Let's wait for @koustuvsinha !!

koustuvsinha:
lets keep the REPL snippets for a later PR. LGTM too :+1:

Rafase282:
I only keep the link back to the original article and make a TOC. I think this is a good format. With the repl in general I think it would be best to leave them for antoher weave. Just as we are spiting articles, later we can focus on adding repls or have someone create them.

PR #807, opened 2016-04-27
Split Article Python Functions

Rafase282:
Splits the article into smaller parts.


theflametrooper:
@alayek please check this :P

Rafase282:
I found an error on the linking back, it should be `Python-Functions` not `Python-Function` Will fix that now.

Python-Functions-Defining.md
```diff
@@ -0,0 +1,25 @@
...
+
+```python
+>>> def fib(n):    # write Fibonacci series up to n
+...     """Print a Fibonacci series up to n."""
+...     a, b = 0, 1
+...     while a < n:
+...         print(a, end=' ')
+...         a, b = b, a+b
+...     print()
+...
```
alayek: I think here we need a REPL snippet. Maybe not in this PR though.

alayek:
# LGTM :+1: 

@koustuvsinha take a look. Hope we are not going too granular!

koustuvsinha:
# LGTM too :+1:

Issue #819, opened 2016-04-29
Create new articles for Python 3 built-in functions


alayek:
We create new articles to serve as quick reference to Python 3 built-in functions, and put them interlinked with the More built-in functions article.

- [ ] `abs()`
- [ ] `all()`
- [ ] `any()`
- [ ] `ascii()`
- [ ] `bin()`
- [ ] `bool()`
- [ ] `bytearray()`
- [ ] `bytes()`
- [ ] `callable()`
- [ ] `chr()`
- [ ] `classmethod()`
- [ ] `compile()`
- [ ] `complex()`
- [ ] `delattr()`
- [ ] `dict()`
- [ ] `dir()`
- [ ] `divmod()`
- [ ] `enumerate()`
- [ ] `eval()`
- [ ] `exec()`
- [ ] `filter()`
- [ ] `float()`
- [ ] `format()`
- [ ] `frozenset()`
- [ ] `getattr()`
- [ ] `globals()`
- [ ] `hasattr()`
- [ ] `hash()`
- [ ] `help()`
- [ ] `hex()`
- [ ] `id()`
- [ ] `input()`
- [ ] `int()`
- [ ] `isinstance()`
- [ ] `issubclass()`
- [ ] `iter()`
- [ ] `len()`
- [ ] `list()`
- [ ] `locals()`
- [ ] `map()`
- [ ] `max()`
- [ ] `memoryview()`
- [ ] `min()`
- [ ] `next()`
- [ ] `object()`
- [ ] `oct()`
- [ ] `open()`
- [ ] `ord()`
- [ ] `pow()`
- [ ] `print()`
- [ ] `property()`
- [ ] `range()`
- [ ] `repr()`
- [ ] `reversed()`
- [ ] `round()`
- [ ] `set()`
- [ ] `setattr()`
- [ ] `slice()`
- [ ] `sorted()`
- [ ] `staticmethod()`
- [ ] `str()`
- [ ] `super()`
- [ ] `tuple()`
- [ ] `type()`
- [ ] `vars()`
- [ ] `zip()`
- [ ] `__import__()`

Please comment below if you are interested. I will block the particular issue accordingly.

Rafase282:
I say we name name `Python-Function-[the function name].md`

meetmangukiya:
I don't see `exit()`

alayek:
@meetmangukiya it's not mentioned [here](https://docs.python.org/3/library/functions.html) as a built-in function.

[...]

PR #826, opened 2016-04-30
Python all function

alayek:
In reference to `all()` in #819

### DIFF
--- a/python-all.md
+++ b/python-all.md
@@ -1,0 +1,40 @@
+# Python all(iterable)
+
+`all()` is a built in function in Python 3, to check if all items of an [_iterable_](https://docs.python.org/3/glossary.html#term-iterable) is `True`. It takes one argument,  `iterable`. 
+
+## Argument
+### iterable
+
+The `iterable` argument is the collection whose all entries are to be checked. It can typically be a `list`, `str`, `dict`, `tuple` etc. 
+
+## Return Value
+The return value would be a boolean. If and only if **all** entries of iterable are `True`, it returns `True`. This function essentially performs a Boolean `AND` operation over all elements.
+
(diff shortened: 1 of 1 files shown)


python-all.md
```diff
@@ -0,0 +1,40 @@
+# Python all(iterable)
+
+`all()` is a built in function in Python 3, to check if all items of an [_iterable_](https://docs.python.org/3/glossary.html#term-iterable) is `True`. It takes one argument,  `iterable`.
```
theflametrooper: "built in function in Python 3": Repetitive use of the word `in`.
Maybe change to: "built in function with Python 3".

alayek: I think it would be better to do it like
> built-in function in Python 3

The _in_ before Python 3 is standard.

python-all.md
```diff
@@ -0,0 +1,40 @@
...
+print(all([9, 8, [1, 2]])) #=> True
+print(all([9, 8, []])) #=> False Because it has []
+print(all([9, 8, [1, 2, []]])) #=> True
+print(all([9, 8, {}])) #=> False Because it has {}
+print(all([9, 8, {'engine': 'Gcloud'}])) #=> True
+
+```
+:rocket: [REPL It!](https://repl.it/CL9U/0)
+
+[Documentation](https://docs.python.org/3/library/functions.html#all)
```
koustuvsinha: Official Docs

python-all.md
```diff
@@ -0,0 +1,40 @@
+# Python all(iterable)
```
koustuvsinha: Python `all(iterable)`

theflametrooper:
Thanks @alayek :shipit:

Rafase282:
The filename is lowercase.

PR #841, opened 2016-05-02
Fix spacing of "PR" in PR Contribute article

erictleung:
No issue reference. Just noticed a lack of spacing in this article randomly and thought I'd make a fix for it 😄  (not sure what happened to the last line of this change though...)


Rafase282:
Thanks for catching that.

PR #860, opened 2016-05-05
Updates Code-of-Conduct.md 

germainep:
adds clauses for trolling and spamming to Code-of-Conduct.md
closes issue #731

PR #861, opened 2016-05-06
shorten the code of conduct and include spamming and trolling

QuincyLarson:
Closes #860


QuincyLarson:
@Rafase282 @germainep here is my significantly shortened version of the code of conduct. I'd love your feedback.

Rafase282:
@QuincyLarson That seems to cover what was needed and provides a chat room instead of different names which is good.

abhisekp:
@QuincyLarson 
# :+1: LGTM

Just that
> *Trolling includes posting inflamitory comments to provoke an emotional response or disrupt discussions.*

We usually have editor wars :fire: (_vim vs atom_), etc. And I don't think that should count as trolling.

----
Trolling is about deceiving or convincing someone into believing or acting in an outrageous or malicious manner and disrupt discussions.

PR #934, opened 2016-05-13
More specific

willnwhite:


qmikew1:
The original wording comes directly from [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Closures). Not sure that changing a couple of words would add clarity, and seems to deviate a bit from what seems to be standard usage of the term 'free variables' in this (and other) context(s).
 > Closures are functions whose inner functions refer to independent (free) variables. In other words, the functions defined in the closure 'remember' the environment in which they were created.

willnwhite:
I couldn't re-commit to this branch so I've opened https://github.com/FreeCodeCamp/wiki/pull/953.

PR #937, opened 2016-05-13
Fix Streak link

pitazo:


Rafase282:
Thanks, I though I had that covered.

PR #940, opened 2016-05-13
Fix BrowniePoints link

pitazo:

### DIFF
--- a/Streaks.md
+++ b/Streaks.md
@@ -1,6 +1,6 @@
 # Streaks
 
-Your user profile displays your recent activity. If you get [Brownie Points](Brownie-Points.md) on several consecutive days, you will start a streak.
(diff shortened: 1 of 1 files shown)


Streaks.md
```diff
@@ -1,6 +1,6 @@
 # Streaks
 
-Your user profile displays your recent activity. If you get [Brownie Points](Brownie-Points/) on several consecutive days, you will start a streak.
+Your user profile displays your recent activity. If you get [Brownie Points](Brownie-Points.md) on several consecutive days, you will start a streak.
```
Rafase282: You don't need to add the `.md` it works fine without it for the wiki. You might need it for it to work on your fork but not on the actual wiki.

pitazo: Oh, I didn't know that.

PR #953, opened 2016-05-14
Added explanation of free variable

willnwhite:


JS-Closures.md
```diff
@@ -1,6 +1,6 @@
 # JavaScript Closures
 
-Closures are functions that refer to independent (free) variables. In other words, the function defined in the closure 'remembers' the environment in which it was created.
+Closures are functions that refer to independent (free) variables (variables that are used locally, but defined in an enclosing scope). In other words, the function defined in the closure 'remembers' the environment in which it was created.
```
Rafase282:
```diff
-Closures are functions that refer to independent (free) variables (variables that are used locally, but defined in an enclosing scope).
+Closures are functions that refer to independent (free) variables, this means variables that are used locally, but defined in an enclosing scope.
```

I think that would be better.

JS-Closures.md
```diff
@@ -1,6 +1,6 @@
 # JavaScript Closures
 
-Closures are functions that refer to independent (free) variables (variables that are used locally, but defined in an enclosing scope). In other words, the function defined in the closure 'remembers' the environment in which it was created.
+Closures are functions that refer to independent (free) variables. Free variables are variables that are used locally, but defined in an enclosing scope. In other words, the function defined in the closure 'remembers' the environment in which it was created.
```
willnwhite: The 'closure' in the last sentence is different to the 'closure' in the first sentence.

Rafase282:
# LGTM

Rafase282:
>In the first sentence, closure is the nested function in the example. In the last sentence, the closure is the outside (higher-order) function. I think the first sentence is correct, but what do you think?

I don't quite see/understand it.

JS-Closures.md
willnwhite: I'm reading Wikipedia now and it seems more like a closure is what's returned by `makeFunc()`:

> closure [refers to] a [function] whose open bindings (free variables) have been closed by (or bound in) the lexical environment, resulting in a closed expression, or closure.

JS-Closures.md
```diff
@@ -1,6 +1,6 @@
 # JavaScript Closures
 
-Closures are functions that refer to independent (free) variables. In other words, the function defined in the closure 'remembers' the environment in which it was created.
+Closures are functions that refer to independent (free) variables. Free variables are variables that are used locally, but defined in an enclosing scope. In other words, the function defined in the closure 'remembers' the environment in which it was created.
```
willnwhite: > Closures are functions that refer to independent (free) variables.

```javascript
var name = "Apple";
function displayName() {
     alert(name);
}
```
This is a function that refers to a free variable, but it is not a closure because it is not returned by an enclosing function. Therefore, I think the first sentence should read:

> Closures are functions, that refer to independent (free) variables, that have been returned by an enclosing function.

Rafase282:
Sorry I haven't go back to you on this. I have been busy and I'll need to familiarize myself with closures first. Try asking at the wiki chat meanwhile.

Rafase282:
I was reading the link and I feel it would be better to expand on this topic.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Closures

willnwhite:
As far as adding an explanation of free variables, I think we're both happy here. I'll open a new PR about the "definition of closure" issue I'm having.

PR #956, opened 2016-05-15
Create Algorithm-Arguments-Optional.md (es.lang)

pitazo:
Title should be in Spanish or English?


Rafase282:
I think they can be in Spanish. I'll have to check.

PR #960, opened 2016-05-15
Alogrithm Confirm The Ending

Oviduzz:


Rafase282:
Thank you for submitting a solution but we already have have an article with solutions for this. No need to create a new article.

Also please read carefully the guidelines for contribution. The filename and content are deviating from our guidelines.

PR #962, opened 2016-05-15
Article: Python Function ID

Gouravchawla:
For issue #819

### DIFF
--- a/Python-Function-ID.md
+++ b/Python-Function-ID.md
@@ -4,14 +4,9 @@
 
 ## Argument
 
-### object
+#### object
 
 The `object` argument can typically be a `int`,`float`, `str`,`list`, `dict`, `tuple` etc.
-
-## Return Value
-
-The return value would be a unique and constant `integer`.
-
 
 ## Code Sample
 
--- a/Python-Function-ID.md
+++ b/Python-Function-ID.md
@@ -1,0 +1,47 @@
+# Python id(object)
+
+`id()` is a built-in function in Python 3, which returns the *identity* of an object. The *identity* is a unique integer for that object during its lifetime. In CPython implementation, this is the address of the object in memory which is not necessarily true for other python implementations.
+
+## Argument
+
+### object
+
+The `object` argument can typically be a `int`,`float`, `str`,`list`, `dict`, `tuple` etc.
+
+## Return Value
+
+The return value would be a unique and constant `integer`.
+
+
+## Code Sample
+
+```python
+a = 2
+print(id(a)) #=> 140454723286976 (Values returned by id() might be different for different users)
+
+b = 3
+print(id(b)) #=> 140454723287008
+
+c = 2
+print(id(c)) #=> 140454723286976 (This is same as id(a) since they both contain the same value and hence have same memory address)
+
+print(id(a) == id(b)) #=> False (since a and b have different values stored in them)
+print(id(a) == id(c)) #=> True (since a and c have same values stored in them)
+
+d = 1.1
+e = 1.1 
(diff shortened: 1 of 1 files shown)


Python-Function-ID.md
```diff
@@ -0,0 +1,47 @@
...
+## Argument
+
+### object
+
+The `object` argument can typically be a `int`,`float`, `str`,`list`, `dict`, `tuple` etc.
+
+## Return Value
+
+The return value would be a unique and constant `integer`.
+
```
theflametrooper: Duplicate line can be removed.

Python-Function-ID.md
```diff
@@ -0,0 +1,47 @@
+# Python id(object)
+
+`id()` is a built-in function in Python 3, which returns the *identity* of an object. The *identity* is a unique integer for that object during its lifetime. In CPython implementation, this is the address of the object in memory which is not necessarily true for other python implementations.
+
+## Argument
+
+### object
```
theflametrooper: Change this to a `h4`

Gouravchawla:
@theflametrooper I've implemented the changes. Could you check if that is what you suggested?

Python-Function-ID.md
```diff
@@ -0,0 +1,42 @@
+# Python id(object)
+
+`id()` is a built-in function in Python 3, which returns the *identity* of an object. The *identity* is a unique integer for that object during its lifetime. In CPython implementation, this is the address of the object in memory which is not necessarily true for other python implementations.
+
+## Argument
+
+#### object
```
Rafase282:
```diff
-#### object
+### object
```

Python-Function-ID.md
```diff
@@ -0,0 +1,42 @@
...
+
+The `object` argument can typically be a `int`,`float`, `str`,`list`, `dict`, `tuple` etc.
+
+## Code Sample
```
Rafase282:
```diff
-## Code Sample
+### Code Sample
```

Rafase282:
I made some changes, there was no reason to go for h4 after a h2.

Gouravchawla:
Thanks @Rafase282 . What else is required for this PR to be accepted?

Rafase282:
From me, just that.

Python-Function-ID.md
```diff
@@ -0,0 +1,42 @@
+# Python id(object)
+
+`id()` is a built-in function in Python 3, which returns the *identity* of an object. The *identity* is a unique integer for that object during its lifetime. In CPython implementation, this is the address of the object in memory which is not necessarily true for other python implementations.
```
alayek: > In CPython implementation

A beginner would want to know what you mean by _CPython implementation_, so better link to the GitHub mirror of CPython interpreter, or some article that talks about this.

alayek: > which is not necessarily true for other python implementations.

I know that CPython is not a perform-ant implementation. PyPy and Pyston would probably handle performance better by moving around objects in memory.

But reading it, it feels like that `id()` need to be same, is not part of the spec. Is that so? I am not fully aware of this.

alayek:
Thanks for the article @Gouravchawla ! Great job :+1: 

I don't come across people often who care about underlying implementation details of a language. Just fix as per the comments on the article, and we are ready to merge.

Gouravchawla:
Thanks @alayek . Also, I've made some changes and I think omitting the details about implementation makes more sense as the id() function is not used that much. Even the official documentation doesn't say much about it.

Rafase282:
# LGTM :+1:

PR #932, opened 2016-05-13
fix #926

akarshHegde:

PR #972, opened 2016-05-17
Fix symbols markdown from #932

theflametrooper:
Addresses https://github.com/FreeCodeCamp/wiki/pull/932/files#r63580359


abhisekp:
Thank you :sparkles: :smiley:

PR #996, opened 2016-05-20
Touches up new D3 article and adds them to the map

Rafase282:
We now have 2 D3 articles, they don't seem to be quite the challenges but I added them to the challeges on the map


theflametrooper:
### **LGTM :+1:**

PR #1000, opened 2016-05-20
Create Python-Function-Range.md

jainaman224:
For issue #819


koustuvsinha:
hi @jainaman224 ! thanks for this awesome PR, will review it soon. However, please file separate PR for different logical changes, like ` US-map-Visulization-using-D3.md` file shouldn't be in this PR, file it in a separate one.

PR #1005, opened 2016-05-21
Use secure images

Rafase282:
Switches imgur images from `http` to `https`

### DIFF
--- a/Adding-Environment-Variables.md
+++ b/Adding-Environment-Variables.md
@@ -6,7 +6,7 @@
 2. Open Varpanel.
 3. In PATH variable, add the relevant path (for example).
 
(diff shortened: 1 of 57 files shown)


shaginthecloud:
Little late for that. All my previous accounts are suspended. Operating system failure. No more access to funds either. Might as well be in third world country. Please no more request for a couple weeks. why does backspace keep copying words. I'm not I'm good condition

Sent from Yahoo Mail on Android 
 
  On Fri, May 20, 2016 at 23:08, Rafael J. Rodriguez<notifications@github.com> wrote:   
Merged #1005.

Rafase282:
What are you talking about @shaginthecloud ?

abhisekp:
thanks for the insight @shaginthecloud . :)

PR #1035, opened 2016-05-25
Move National Harbor from WA to MD

denmch:
National Harbor is in Ft. Washington, MD, not in Washington state.


hallaathrad:
Thanks, @denmch!

PR #1053, opened 2016-05-28
Markdown fixes for hash tables

Rafase282:
## Changes:
- Add space between the list and the text before it as we do in all other articles. (It gets displayed this way which is why I enforce that structure)
- Changed the first h2 to h3 so the chat will display some information instead of just the title.
- Made list on the first level instead of second level of indentation.


abhisekp:
Looks perfecto :+1: 
<h1 align="center"> :sparkles:  :smile_cat:  :+1: LGTM :smiley: :+1:   :sparkles:  </h1>

PR #1091, opened 2016-06-07
Update algorithm truncate

bmorelli25:
Added in an advanced one-line solution that uses nested ternary operators. The solution also utilizes the slice() method - which is what the student is "supposed" to use to solve this Algorithm challenge. The current solution uses the substr() method, so I added that into the hint section as well.

### DIFF
--- a/Algorithm-Truncate-A-String.md
+++ b/Algorithm-Truncate-A-String.md
@@ -1,20 +1,32 @@
 # Algorithm Truncate a String
 
-### Explanation:
+:triangular_flag_on_post: Remember to use [**`Read-Search-Ask`**](FreeCodeCamp-Get-Help) if you get stuck. Try to pair program :busts_in_silhouette: and write your own code :pencil:
+
+### :checkered_flag: Problem Explanation:
 
 We need to reduce the length of the string or **truncate** it if it is longer than the given maximum lengths specified and add `...` to the end. If it is not that long then we keep it as is.
 
-## Hint: 1
+#### Relevant Links
+
+- [String.prototype.slice()](https://github.com/FreeCodeCamp/FreeCodeCamp/wiki/js-array-prototype-slice)
+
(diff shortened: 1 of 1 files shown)


Rafase282:
As I discussed in the public chat yesterday, I don't like the idea of nesting ternary operators because it can be a pain to read the code, being a one line does not make it always better.

```js
function truncateString(str, num) {
  // Clear out that junk in your trunk

  if (str.length <= num) {
    return str;
  } else {
    return str.slice(0, num > 3 ? num - 3 : num) + '...';
  }
}

truncateString("A-tisket a-tasket A green and yellow basket", 11);
```

This is the solution I decided on sticking.

bmorelli25:
Oh it certainly is a pain to read - just figured it would be good for students to get experience with learning how to read them so they can figure out what in the heck is going on if they ever see it again.

But, that's fair. I'll swap the code with your recommendation above so we get the best of both worlds.

Rafase282:
@bmorelli25 would you mind trying to see if we [our slice article](https://github.com/FreeCodeCamp/FreeCodeCamp/wiki/JS-Array-Prototype-Slice) instead of the link you provided?

Also I will need you to update the whole article as stated on #1094 and leave only our names on the thank section. The script would be on the advance category.

Please add this as the basic one:

```js

function truncateString(str, num) {
  // Clear out that junk in your trunk
  if (str.length > num && num > 3) {
    return str.slice(0, (num - 3)) + '...';
  } else if (str.length > num && num <= 3) {
    return str.slice(0, num) + '...';
  } else {
    return str;
  }

}

truncateString("A-tisket a-tasket A green and yellow basket", 11);
```

bmorelli25:
Will do

bmorelli25:
Committed and ready for feedback!

Rafase282:
Thanks, great work.

Issue #1094, opened 2016-06-07
Update All Algorithms Challenges


Rafase282:
# What we need:

We need all articles to follow this [template](https://github.com/FreeCodeCamp/FreeCodeCamp/wiki/Wiki-Template-Challenge-Solution).

## What to watch out for:
There should be no more than three solutions, per challenge, we have to make sure that we provide explanations, and resources via links, either articles about built in functions that would be needed, algorithms, or anything related that can or was used to solve the problem.

## How to get started

1. Check for available articles, this means make sure no one is working on them already.
2. Ask to be assigned to any of the members on the **Assignees** list to the right. Non FCC team members cannot be added to this list, but we can add your name to it.
3. Once you get an assignment, let us know the link to the pull request, it will be added next to the challenge name along with your name.

## Adding links

First check the FCC challenge, they have general links used as hints, note them for those are the ones we must include. However, if we have our own resources [here](https://github.com/FreeCodeCamp/FreeCodeCamp/wiki) then use that instead. You do this by searching for the article. for example if you are looking for slice:

![search](https://i.imgur.com/NU4yJ1m.png)

Then you canuse that instead of an external resource. The template uses 3 links, but it can be fewer or more as needed.

When looking for links for the solutions, check which built in functions or specific algorithm is used and search our wiki for related articles, if none is found, only then use an external resource.

**Do not repeat links** 
If you already use a link in the general section or one of the solutions, then do not add it to another one in the same article.

## Adding code explanations

You try to explain it with your own words but not in a personal voice.

**Only 3 solutions are accepted**
If there are more than three then talk with us about which one to remove or merge at the [chat](https://gitter.im/FreeCodeCamp/Wiki).

> We don't need many of them and a maximum of three solutions will be accepted in the following categories:

>Basic
Intermediate
Advance
One code solution per category. If we already have enough solutions and yours is similar to one we already have but you strongly feel like yours will bring something important then we encourage you to merge your solution with the similar one we already have and expand or write a detailed explanation of the code.

## Challenges to be updated:

- [x] [Arguments Optional](https://github.com/FreeCodeCamp/wiki/blob/master/Algorithm-Arguments-Optional.md) (@Rafase282) #1093 
- [x] [Binary Agents](https://github.com/FreeCodeCamp/wiki/blob/master/Algorithm-Binary-Agents.md) (@atjonathan) #1099 
[...]

PR #1097, opened 2016-06-07
Updated Algorithm Reverse A String

atjonathan:
Refer #1094


Rafase282:
Make the links without the inline code quote and then it should be fine.

PR #1115, opened 2016-06-10
Update where do I belong

alayek:
Refer #1094

### DIFF
--- a/Algorithm-Where-Do-I-Belong.md
+++ b/Algorithm-Where-Do-I-Belong.md
@@ -1,20 +1,31 @@
 # Algorithm Where do I Belong
 
-### Explanation:
+:triangular_flag_on_post: Remember to use [**`Read-Search-Ask`**](FreeCodeCamp-Get-Help) if you get stuck. Try to pair program :busts_in_silhouette: and write your own code :pencil:
+
+### :checkered_flag: Problem Explanation:
 
 This can be a tricky problem to understand. You need to find where in the array a number should be inserted by order, and return the index where it should go.
 
-## Hint: 1
+#### Relevant Links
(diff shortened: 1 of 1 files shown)


Rafase282:
It is only missing links on the solution's sections.

Rafase282:
Please add links to functions that are used on the solutions.

Rafase282:
I was working on this and I came up with a shorter version of the second solution but decided that it was wrong because the challenge ask us to find the index after we sort from smaller to larger where the number would go, but it does not ask us to actually insert it so technically the second solution is wrong.

Please remove it and remove the second name. You can add this one:

```js
function getIndexToIns(arr, num) {
  // Find my place in this sorted array.
  function minToMax(arr){
    return arr.sort((curr,next)=>curr>next);
  }
  var index = minToMax(arr)  
    .findIndex(function(n){
      return num <= n;
    });
  return index === -1? arr.length : index;
}

getIndexToIns([40, 60], 500);
```

I'm still working on refining it but while I could make it to stand a fewer lines, I think modulizing it like this is more future proof in real life than cramming all. Any thoughts on this?

alayek:
Yes, the second solution alters the original array, while your solution does not. Adhering to the the principles of pure functions, this is a much better solution. Because it doesn't alter anything.

I will remove the second one and include yours, once you are done refining.

Rafase282:
@alayek currently the only idea that comes to mind is to inline the sorting part instead of making it a function, what do you think of it? Worth it or should I leave it as is?

alayek:
@Rafase282 sure, if you inline `return arr.sort((curr, next) => curr > next);`, it doesn't hurt readability.

PR #1132, opened 2016-06-13
Updates steamroller and algorithm template

Rafase282:
Refers #1094

I removed the credit for the second solution as that person basically copied the answer from Stackoverflow. The person did give credit on the link but I still feel the name should be removed as it was not something that he came on his own.

Also I updated the thanks section of the template and all articles so far.


atjonathan:
### **LGTM :+1:**

PR #1135, opened 2016-06-13
Updated Validate US Telephone Numbers

atjonathan:
Refers #1094


Rafase282:
For the second one, please add the library to the links section and remove the repeated link. You coudl switch it to the Regex.test() from MDN.

PR #1140, opened 2016-06-14
Update Algorithm Pig Latin

dakshshah96:
Refer #1094


Rafase282:
Come to the wiki chat, I'll help you figure out the problem.

PR #1141, opened 2016-06-14
Update Algorithm Pig Latin

dakshshah96:
Refer #1094

### DIFF
--- a/Algorithm-Pig-Latin.md
+++ b/Algorithm-Pig-Latin.md
@@ -1,20 +1,34 @@
 # Algorithm Pig Latin
 
-### Problem Explanation:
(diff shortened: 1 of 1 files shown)


Algorithm-Pig-Latin.md
```diff
@@ -117,40 +157,33 @@ translatePigLatin("consonant");
 
 :rocket: [Run Code](https://repl.it/CLmv/0)
 
-### Fourth Solution (recursive and declarative)
```
Rafase282: I would prefer this answer over any of the others, the other 3 uses regex, this is the only one that does not.

dakshshah96: True! Alright, I'll make the necessary changes. Should I put this as the Basic solution?

Rafase282:
I would say drop the second solution from the original so they would go in this order, 1,3,4 as beginer, intermediate, advance.

dakshshah96:
@Rafase282 I've implemented the suggestions. Hope it's alright!

Rafase282:
Great, thanks!

PR #1163, opened 2016-06-20
<Added solution using a while loop>

zk433:


alayek:
@zk433 thanks for your interest in contributing. 

However, this solution is no different from `for` loop. Any `while` loop can be converted to a `for` loop, and vice-versa.

Also, we have the rule about not accepting more than three solutions to any algorithm challenges. We have no choice but to close the pull request.

---

I am really sorry, but there are other places in our Wiki you could contribute. Please understand we really appreciate your enthusiasm in contributing and helping others.

Do talk to us in our [Wiki room](https://gitter.im/FreeCodeCamp/Wiki). We have tons of things for you to contribute to!

PR #1164, opened 2016-06-20
Update Algorithm Symmetric Difference

dakshshah96:
Refer #1094 
- Removed one of the solutions (https://repl.it/CLod/0) because it wasn't working.
- Added a new Basic solution.

### DIFF
--- a/Algorithm-Symmetric-Difference.md
+++ b/Algorithm-Symmetric-Difference.md
@@ -1,44 +1,41 @@
 # Algorithm Symmetric Difference
 
-### Problem Explanation:
+:triangular_flag_on_post: Remember to use [**`Read-Search-Ask`**](FreeCodeCamp-Get-Help) if you get stuck. Try to pair program :busts_in_silhouette: and write your own code :pencil:
 
-Symmetric Difference is the difference between **two** sets.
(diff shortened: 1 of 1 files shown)


Rafase282:
The intermediate solution is mine, if you added the basic then you can add your name, which means we can remove the other names.

Also https://repl.it/CLod/0 works so we don't need to remove this solution.

dakshshah96:
Can you provide the explanation for https://repl.it/CLod/0 ?
And about the names, if we include  https://repl.it/CLod/0, which names should I keep?

Rafase282:
Names to keep:

- @Rafase282
- @imranismail
Yours if you still add the solution. 

The others can be removed.

Rafase282:
Links for the ES6 solution:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/has
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_operator
map, reduce and filter we should have on FCC.

atjonathan:
### **LGTM :+1:**

PR #1192, opened 2016-06-28
fix broken links

profoundhub:
Fixed Broken Links


Rafase282:
They are not broken, they are just not meant to be accessible from this repo.

PR #1196, opened 2016-06-30
new section: create your own forEach

cliffordfajardo:
added this section for people who are curious about how forEach works behind the scenes. It's not magic, it's javascript =).

### DIFF
--- a/JS-Array-Prototype-ForEach.md
+++ b/JS-Array-Prototype-ForEach.md
@@ -52,3 +52,18 @@ numbers.forEach(
 document.write(sum);
 // Output: 33
 ```
+
+## How to create your own forEach for learning purposes
+```javascript
+function forEach(array, callback){
+  for(let i = 0; i < array.length; i++){
+    //apply the passed in callback on each element in the array
+    callback(array[i]);
+  }
+}
+function logValue(value, index){
+  console.log(`Element is: ${value}` )
+}
+forEach([1,2,3], logValue); //'logs Element is: 1' .... 'Element is: 2 ......'Element is: 3'
+
+```


JS-Array-Prototype-ForEach.md
```diff
@@ -52,3 +52,18 @@ numbers.forEach(
...
+    //apply the passed in callback on each element in the array
+    callback(array[i]);
+  }
+}
+function logValue(value, index){
+  console.log(`Element is: ${value}` )
+}
+forEach([1,2,3], logValue); //'logs Element is: 1' .... 'Element is: 2 ......'Element is: 3'
+
+```
```
atjonathan: Add a [repl](https://repl.it) snippet.

JS-Array-Prototype-ForEach.md
```diff
@@ -52,3 +52,18 @@ numbers.forEach(
 document.write(sum);
 // Output: 33
 ```
+
+## How to create your own forEach for learning purposes
```
atjonathan: Add a line underneath this one.

JS-Array-Prototype-ForEach.md
```diff
@@ -52,3 +52,18 @@ numbers.forEach(
...
+```javascript
+function forEach(array, callback){
+  for(let i = 0; i < array.length; i++){
+    //apply the passed in callback on each element in the array
+    callback(array[i]);
+  }
+}
+function logValue(value, index){
+  console.log(`Element is: ${value}` )
```
alayek: Why do you need `${}` syntax?

atjonathan: @alayek, those are template literals, introduced in ES6.

cliffordfajardo: ES6 string interpolation:`console.log(`Element is: ${value}` )`
Pre-ES6 string interpolation: `console.log("Element is:, value")

When you have more than 1 variable you want included in the log, I believe it makes for sense to do ES6 string interpolation. I also find it easier to read. through.

```js
function logValue(value, index){
  console.log(`Element is: ${value} and the index is: ${index}` )
```
instead of something like:
```js
function logValue(value, index){
  console.log(`Element is: " + value +  "and the index is: " +  index )
```

alayek: That's ok. I also noticed the usage of `let`.

---

I am not against using ES6. But we don't teach ES6 syntax in our core curriculum. Even, the codemirror editor we use, gives syntax error when we use arrow functions etc. from ES6.

The writer should at the very least mention that this is in ES6 syntax. Or simply use ES5 syntax, because target audience would be most familiar with it.

alayek:
`forEach` is a higher order function whose argument function takes other arguments as well.

It takes each element, its index, and entire array. Please extend your implementation to handle those as well.

I like the pointfree style you are using, but if you are really implementing it for your own usage (say, polyfill), go all the way. Use this syntax: 

```js
Array.prototype.forEach = Array.prototype.forEach || function(element, index, array){ 
/* implementation with for loop */
}
```

We should also mention somewhere that this is not what the real implementation is. The real implementation, at least in V8, is much more complex. [This](https://news.ycombinator.com/item?id=7937321) might interest you.

cliffordfajardo:
@alayek -  I was also thinking about showing how to add it to the Array prototype, but for beginners who don't understand what prototype is that would be confusing. I'll update this example & the other pull requests to add to the array prototype. I'll also include a description that this is a pseudo implementation of how it works. Thanks for the link

Rafase282:
@cliffordfajardo please create the articles at the [forum](http://forum.freecodecamp.com/) from now on.

PR #1198, opened 2016-06-30
new section showing map function's implementation

cliffordfajardo:
this addition is simply for learners who would like to understand how the map function works behind the scenes. This is one of many ways to implement it. It's not magic, it's javascript =).

### DIFF
--- a/JS-Array-Prototype-Map.md
+++ b/JS-Array-Prototype-Map.md
@@ -70,3 +70,20 @@ var jediMastersSayHi = jediMasters.map(sayHi); // We're done mapping, now consol
 ```
 
 We can use map in strings and other types of objects if we use `Array.prototype.call(objectToBeMapped, callbackFunction)`, for more examples, go look other sites and practise it.
+
+
+```javascript
+function map(array, callback){
+  let results = [];
+  for(let i = 0; i < array.length; i++){
+    //apply the passed in callback on each element in the array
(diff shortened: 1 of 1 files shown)


JS-Array-Prototype-Map.md
```diff
@@ -70,3 +70,20 @@ var jediMastersSayHi = jediMasters.map(sayHi); // We're done mapping, now consol
 ```
 
 We can use map in strings and other types of objects if we use `Array.prototype.call(objectToBeMapped, callbackFunction)`, for more examples, go look other sites and practise it.
+
```
atjonathan: Remove this line.

JS-Array-Prototype-Map.md
```diff
@@ -70,3 +70,20 @@ var jediMastersSayHi = jediMasters.map(sayHi); // We're done mapping, now consol
...
+    results.push(callback(array[i]));
+  }
+  return results;
+}
+function doubleNumber(number){
+  return number * 2;
+}
+
+let doubledNumbersArray = map([1,2,3], doubleNumber); //[2,4,6]
+```
```
atjonathan: :rocket: Add a [repl](https://repl.it) snippet, **`:rocket: [Run Code](url)`**

Rafase282:
@cliffordfajardo please create the articles at the [forum](http://forum.freecodecamp.com/) from now on.

PR #1207, opened 2016-07-02
Update Algorithm Chunky Monkey

### LINKED ISSUE #1205: Chunky Monkey needs better solutions
Referring https://github.com/FreeCodeCamp/FreeCodeCamp/issues/9500
> The basic solution is very convoluted and intermediate isn't much better. Advanced is OK.
>
> Challenge also might benefit from mentioning the <code>splice</code> method since that is a much simpler approach than <code>slice</code>.
>
> Propose eliminating that ugly basic solution, demoting intermediate to basic, and advanced to intermediate. Also recommend adding <code>splice</code> to list of suggested methods.

dakshshah96:
Closes #1205.
- Add new **Basic** solution which uses the `splice()` method.
- Add simplified **Intermediate** solution which uses the `slice()` method.

@raisedadead: What do you think of this?

@Rafase282: Regarding the names in **Credits**, are there any I should remove?

### DIFF
--- a/Algorithm-Chunky-Monkey.md
+++ b/Algorithm-Chunky-Monkey.md
@@ -1,12 +1,10 @@
 # Algorithm Chunky Monkey
 
-![](https://i.imgur.com/RdrnNvA.png)
-
 :triangular_flag_on_post: Remember to use [**`Read-Search-Ask`**](FreeCodeCamp-Get-Help) if you get stuck. Try to pair program :busts_in_silhouette: and write your own code :pencil:
 
 ### :checkered_flag: Problem Explanation:
 
-Our goal for this Algorithm is to split `arr` (first argument) into smaller chunks of arrays with the length provided by `size` (second argument). There are 4 green checks (objectives) our code needs to pass in order to complete this Algorithm:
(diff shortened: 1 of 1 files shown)


atjonathan:
### **LGTM :+1:**

Rafase282:
Personally I think there is no much difference between the old intermediate and the new basic.

From the previous ones, the basic was the only that was different and used basic functions and iterators, the others use slice and splice that while are not the same does similar thigns and overall the code was similar.

I vote for keeping the old basic.

@dakshshah96 as for the names, these challenges have already been moved to the forum where the whole credit section has been removed as the new standard.

atjonathan:
I agree with @Rafase282 with further scanning.

alayek:
In general, I would recommend against `splice()` method. It alters or mutates the original array. Does not adhere to good practices of functional programming, namely, writing pure functions that always produce same outputs for same inputs.

Always use `slice()` over `splice()` when you can. Same goes for `concat()` vs. `push()`.

Rafase282:
That's a good point, I didn't check that. In some challenges we are not supposed to modify the original array.

PR #1209, opened 2016-07-02
Forum

Rafase282:
My current progress on the migration


Rafase282:
I think the conflict is due to the new articles that were not here.

Rafase282:
Done

atjonathan:
@Rafase282, let's work on this new branch instead of the one on your fork.

Rafase282:
Alright, just make sure to check if the name is linked anywhere else and and if it is, to change the links with the full link of the forum post, the good thing is that even if the title change, the link will still work so we don't have to worry about links after they are set, they will always point tot he right article

Rafase282:
# New Changes:

#1210

Rafase282:
Well is no much reviewing as to it is to let everyone know the new changes. I used this because the conversation was locked and it was the first so I was thinking of having this as the center but perhaps I'll just make an issue instead.

PR #1210, opened 2016-07-02
Covers all 'Challenge-A' articles

Rafase282:
These articles are already on the forum

Refers #1211


alayek:
# LGTM :+1:  

Let's :shipit: :package:

PR #1218, opened 2016-07-03
Adds 'advanced' algorithm solution to challenge

### LINKED ISSUE #1217: Add advanced solution to 'Pairwise' challenge solution
#### Pairwise
https://www.freecodecamp.com/wiki/en/algorithm-pairwise/

#### This challenge is missing an 'advanced' solution
I believe that my solution for this challenge represents a suitably more 'advanced solution' for this challenge, including avoiding imperative for loops and using declarative statements using native array functions

I don't want to clone the whole wiki locally and make a pull request, so how do I contribute / add this to the solution?

#### Your Code
This solution uses only native array operations and avoids for loops. 
```
function pairwise(arr, arg) {
  // search array for elements that when paired, equal the second argument, then sum their indices
  
  return arr.reduce( function (a,b,index){ // use native reduce to collect running total of summed indices
      var search = arg - b; // get difference of current item so we know what value will sum to arg

      // check if search value in rest of the array, but also make sure it doesn't match current search index
      if ( arr.indexOf(search) != -1 && arr.indexOf(search) != index ){ 
         var total = index + arr.indexOf(search);  // if found, add to the runnning total
         arr.splice(index,1,NaN); // remove current index from the array
         arr.splice(arr.indexOf(search),1,NaN); // remove the other matched element from the array
         return a + total; //return the running total back to reduce for next item
      }
      return a; // simply return previous total if no operations needed
  },0);
  
}
```

Code explanation:
See comments in code
This code takes advantage of the fact that the native Array.prototype.indexOf() method will return the *lowest* index of the value it finds, a requirement of the challenge. Given that you start with the first item in the array (automatically the lowest of it's value), you're guaranteed to always find the lowest pairs, before removing them from the search space.

### Username to add:
@geligelu

ortonomy:
closes #1217

### DIFF
--- a/Algorithm-Pairwise.md
+++ b/Algorithm-Pairwise.md
@@ -152,7 +152,7 @@ pairwise([1,4,2,3,0,5], 7);
 
 ## :rotating_light: Advanced Code Solution:
 
-````
+````javascript
 function pairwise(arr, arg) {
   // search array for elements that when paired, equal the second argument, then sum their indices
   // make a local copy of the arguments object so we don't modify it directly
@@ -176,7 +176,7 @@ function pairwise(arr, arg) {
 pairwise([1,4,2,3,0,5], 7);
 ````
 
-:rocket: [Run Code](https://repl.it/CLpC/1)
+:rocket: [Run Code](https://repl.it/CLpC/2)
 
 ### Code Explanation:
 See comments in code
--- a/Algorithm-Pairwise.md
+++ b/Algorithm-Pairwise.md
@@ -45,7 +45,7 @@ function pairwise(arr, arg) {
  // Set sum of indices to zero
  var sum = 0;
  // make a local copy of the arguments object so we don't modify it directly
- var pairArr = arr;
+ var pairArr = arr.slice();
  // looping from first element
  for(i = 0; i < pairArr.length; i++) {
    //Looping from second element by setting first element  constant
@@ -156,7 +156,7 @@ pairwise([1,4,2,3,0,5], 7);
 function pairwise(arr, arg) {
   // search array for elements that when paired, equal the second argument, then sum their indices
   // make a local copy of the arguments object so we don't modify it directly
-  var pairArr = arr;
+  var pairArr = arr.slice();
   return pairArr.reduce( function (a,b,index){ // use native reduce to collect running total of summed indices
       var search = arg - b; // get difference of current item so we know what value will sum to arg
 
--- a/Algorithm-Pairwise.md
+++ b/Algorithm-Pairwise.md
@@ -44,16 +44,18 @@ It is easy to confuse indices as being numbers, but since you will be interactin
 function pairwise(arr, arg) {
  // Set sum of indices to zero
  var sum = 0;
+ // make a local copy of the arguments object so we don't modify it directly
+ var pairArr = arr;
  // looping from first element
- for(i = 0; i < arr.length; i++) {
+ for(i = 0; i < pairArr.length; i++) {
    //Looping from second element by setting first element  constant
-   for(j = i + 1; j < arr.length; j++) {
+   for(j = i + 1; j < pairArr.length; j++) {
      // Check whether the sum is equal to arg
-     if(arr[i] + arr[j] == arg) {
+     if(pairArr[i] + pairArr[j] == arg) {
        //Add the indices
        sum += i + j;
        //Set the indices to NaN so that they can't be used in next iteration
-       arr[i] = arr[j] = NaN;
+       pairArr[i] = pairArr[j] = NaN;
      }
    }
  }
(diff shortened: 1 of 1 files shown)


alayek:
This solution is correct, but I have a concern about this:

This uses `splice()`, which mutates the array. Side effects lead to bugs. Especially when you are calling a `reduce` on the very same array, and mutating the array on the fly.

Please use `slice()` instead of `splice()`.

geligelu:
@alayek - this is an interesting point, but using splice was entirely the point. I want the 'used' numbers to be eliminated from the array without having to do complicated checks or use an array to keep track of the 'used' indices.  So this is not a side-effect as opposed to intentionally mutating the array as reduce goes through the loop.

Secondarily to this point the first  "basic" solution also modifies the source array on the fly, as evidenced by this code section:

````
for(i = 0; i < arr.length; i++) {
   //Looping from second element by setting first element  constant
   for(j = i + 1; j < arr.length; j++) {
     // Check whether the sum is equal to arg
     if(arr[i] + arr[j] == arg) {
       //Add the indices
       sum += i + j;
       //Set the indices to NaN so that they can't be used in next iteration
       arr[i] = arr[j] = NaN;
     }
   }
 }
```` 
If I remove it from this solution, then it should be removed from the basic code solution too?

**Comment by alayek** [comment]
Yes, please update your PR with that change in both those solutions. What I would recommend is creating a new array from the array `arr` passed as argument; and modify it as you see fit.

**Comment by geligelu** [comment]
@alayek - thanks for the feedback. To confirm, I will not change the algorithm for the advanced solution, but changes to be made are:
1. Basic solution to avoid mutating the original 'arr' argument object
2. Create a copy into a new array variable from the original 'arr' argument object in the new advanced solution.

**Comment by geligelu** [comment]
@alayek - made a new commit, pls chk. Do I need to make a new pull request?

File: Algorithm-Pairwise.md [hunk 1]
```diff
@@ -44,16 +44,18 @@ It is easy to confuse indices as being numbers, but since you will be interactin
 function pairwise(arr, arg) {
  // Set sum of indices to zero
  var sum = 0;
+ // make a local copy of the arguments object so we don't modify it directly
+ var pairArr = arr;
```
alayek: This still mutates the original array. You are creating a new reference pointer that still points to the same array as `arr`.

You need to create a new Array `pairArr`, using the values in `arr`. Look into `Array.prototype.from()` or `Array.prototype.slice()`.

Algorithm-Pairwise.md
```diff
@@ -148,11 +150,52 @@ pairwise([1,4,2,3,0,5], 7);
 - [JS Array Prototype IndexOf](JS-Array-Prototype-IndexOf)
 - [JS Array Prototype Push](JS-Array-Prototype-Push)
 
+## :rotating_light: Advanced Code Solution:
+
+````
+function pairwise(arr, arg) {
+  // search array for elements that when paired, equal the second argument, then sum their indices
+  // make a local copy of the arguments object so we don't modify it directly
+  var pairArr = arr;
```
alayek: Same as [above](#r69389208)

geligelu:
@alayek  - implemented local var using ``arr.slice()`` to make a copy. ``arr.from()`` is ES6, so won't use that here.

geligelu:
please squash the commits too.

Algorithm-Pairwise.md
```diff
@@ -148,11 +150,52 @@ pairwise([1,4,2,3,0,5], 7);
 - [JS Array Prototype IndexOf](JS-Array-Prototype-IndexOf)
 - [JS Array Prototype Push](JS-Array-Prototype-Push)
 
+## :rotating_light: Advanced Code Solution:
+
+````
```
Rafase282:
```javascript

This will make it easier to read.

alayek:
# LGTM :+1:  

Let's :shipit: :package:

Rafase282:
Lets wait for the javascript declaration on the code first.

geligelu:
@Rafase282  - sorry, what does 'javascript declaration' mean?

Rafase282:
basically the comment I did on the code, do 

**```javascript**

instead of just  **```**

geligelu:
@Rafase282  - sorry. Didn't see the comment. Adding.

geligelu:
@Rafase282 - declaration added.

geligelu:
First PR complete. 🎉

Rafase282:
I found something the run code at the bottom is an extra. I'll remove it. Thanks for the solution.

geligelu:
Doh! Should have checked more. Hoping to make more contributions in the future!

geligelu:
@Rafase282 @alayek  - out of curiosity, when is the code released to live? Do you have a schedule?

Rafase282:
It is already, just not on the wiki from the nav bar yet. it is even on the forum.

PR #1221, opened 2016-07-03
Fix broken links

studenton:


studenton:
@Rafase282 Why wasn't it merged? Can I know?

Rafase282:
Yes, see here https://github.com/FreeCodeCamp/FreeCodeCamp/issues/9513#issuecomment-230159736

Rafase282:
In this particular case, the links are meant to be relative and the way they are. they will not work if you click them in this forum but they will work if you click them on the react wiki or the github one as intended.

If you want to help with links I recommend you take a look a this issue https://github.com/FreeCodeCamp/FreeCodeCamp/issues/9513#issuecomment-230159736

studenton:
Got it. Thanks for explaining.

PR #1224, opened 2016-07-07
Create Wiki About Logic Gate-AND

ShailBala:


alayek:
@ShailBala thanks for your contribution. But I would like to know what should be the context of this article.

For instance, we have articles on JS or Python or Java, and people might need them because they are all used in the web - back-end or front-end. However, at this moment, we don't have campers who would need to know or use low-level (hardware or transistor) electronic concepts like this.

---

I would recommend
- Changing file name to categorize these better ( @Rafase282 FYI)
- Put some code examples in JS or Python or Java or any other Turing complete language which shows binary operation like this.
- Mention some usages.

PR #1225, opened 2016-07-07
Created Data-Structure-Stacks.md

varunu28:


varunu28:
@alayek I have created the article for Stack data structure. Please suggest the changes which are needed to be done.

Data-Structure-Stacks.md
```diff
@@ -0,0 +1,60 @@
+# Data Structure Stacks
+
+`Stack` is a linear data structure which follows a particular order in which the operations are performed. The order may be 
+LIFO(Last In First Out) or FILO(First In Last Out).
```
alayek: I have never encountered the term `FILO` in standard text. I only know of `LIFO`. Can you cite some reference to this? Otherwise, remove it please.

Data-Structure-Stacks.md
```diff
@@ -0,0 +1,60 @@
...
+
+The basic three operations which can be performed on a `Stack`
+
+1. `Push`: Adds an item in the `Stack`. If the stack is full, then it is said to be an Overflow condition. 
+2. `Pop`: Removes an item from the `Stack`. The items are popped in the reversed order in which they are pushed. If the `Stack` is empty, 
+           then it is said to be an Underflow condition. 
+3. `Peek`: Gets the topmost item of the `Stack`.
```
alayek: Emphasize that it does not remove the item. Btw, is there some way to get the length or height if the stack? If so, maybe you should mention that as an operation too. Even if it's an auxiliary operation, and not a basic operation specific to Stack.

Data-Structure-Stacks.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Real life example of a `Stack`
+
+Consider a number of plates kept one upon the one another on a dining table. Now the first plate which we have kept would be the last plate
```
alayek: Can you share an image from Google image (with proper credits, of course).

Data-Structure-Stacks.md
```diff
@@ -0,0 +1,60 @@
...
+to be picked up which explains the First In Last Out. In a similar way the plate which was last to go on the top of the collection
```
alayek:
```diff
- First In Last Out
+ Last In First Out
```

alayek:
@varunu28 left some comments. Please check. Other than that, LGTM from my side.

Data-Structure-Stacks.md
varunu28: I didn't understood this comment. What is this regarding?

Data-Structure-Stacks.md
varunu28: Added

Data-Structure-Stacks.md
varunu28: Done the change. I have mentioned the size function for stack which gives the length of the stack

Data-Structure-Stacks.md
varunu28: I also didn't found any code examples for this. But this concept is applicable to the stack when we keep popping elements from the stack then the last element which we would get will the element which was first pushed to the stack.

Data-Structure-Stacks.md
```diff
@@ -0,0 +1,63 @@
+# Data Structure Stacks
+
+`Stack` is a linear data structure which follows a particular order in which the operations are performed. The order may be 
+LIFO(Last In First Out) or FILO(First In Last Out).
+
+The basic three operations which can be performed on a `Stack`
```
atjonathan: ~~on a `Stack`~~ on a `Stack`:

Data-Structure-Stacks.md
```diff
@@ -0,0 +1,63 @@
...
+
+1. `Push`: Adds an item in the `Stack`. If the stack is full, then it is said to be an Overflow condition. 
+2. `Pop`: Removes an item from the `Stack`. The items are popped in the reversed order in which they are pushed. If the `Stack` is empty, 
+           then it is said to be an Underflow condition. 
+3. `Peek`: Gets the topmost item of the `Stack`. It does not removes the item from the stack.
+
+## Real life example of a `Stack`
+
+Consider a number of plates kept one upon the one another on a dining table. Now the first plate which we have kept would be the last plate
```
atjonathan: This doesn't make much sense. It could be re-factored. I'm not sure what you mean, so can't give guidance.

varunu28: Can you be a bit more precise?

Data-Structure-Stacks.md
varunu28: Done

Rafase282:
@varunu28 please create the articles at the [forum](http://forum.freecodecamp.com/) from now on.

PR #1236, opened 2016-07-12
Advanced code solution not working for repeat a string algorithm

chakrihacker:
Algorithm Repeat a string solution advanced code section removed


atjonathan:
Please use the [forum](http://forum.freecodecamp.com/) to update wiki articles.

Rafase282:
@chakrihacker I found out the problem, it was when the program was renamed, I already fixed it on the forum.

PR #1249, opened 2016-08-12
Added var-ci to autoclose issues and pull requests accordingly

atjonathan:


.varci.yml
```diff
@@ -0,0 +1,26 @@
...
+
+      Happy Coding!
+    close: true
+  comment_and_close_issues:
+    name: "Comment, label & close issues"
+    events: [ issues ]
+    label: invalid
+    when:
+      - action = "opened" or action = "reopened"
+      - title matches "/request|fix|create|/i" or body matches "/request|fix|create|/i"
```
BerkeleyTrue: Do we need to look for specific titles here? Shouldn't all the issues be closed?

atjonathan: Yeah actually I will fix this.

atjonathan:
@BerkeleyTrue this should be good to go :smiley:

PR #1253, opened 2016-08-19
update/Algorithm-Steamroller.md

pro-mos:
Added Advanced code solution for the Steamroller algorithm.


Rafase282:
Please add it to the forum once you are level 1.

PR #1254, opened 2016-08-19
Forum

Rafase282:
this should get closed automatically.


atjonathan:
Yeah it needs fixed. When it's fixed it will only trigger on new pull requests so we can wait till we fix the configuration before making another test :+1:

PR #1259, opened 2016-08-21
Update Algorithms-Binary-Search.md

ilogin:
Added explanation for why the time complexity of a binary search is O(log(n))


Rafase282:
We are not accepting changes here. Please visit the forum.

PR #1261, opened 2016-08-28
Updated Map wiki page to fix Algorithm-Where-Art-Thou link

wjhurley:
The Algorithm-Where-Art-Thou pointed to the wiki page
Algorithm-Where-Art-Thou, which does not exist. The correct wiki page is
Algorithm-Wherefore-Art-Thou, so I updated the text for the link to
match the name of the challenge as it appears on FCC's website and
updated the link as well to point to the existing page.


atjonathan:
We have moved to the [forum](http://forum.freecodecamp.com/). However, I will merge this as it breaks functionality within this wiki.

